PR #105, opened 2015-02-16
Removing the erroneous example

mikeSimonson:
Getting the EntityManager in the postUp method is flawed: if you have several unexecuted migrations in the project, the EntityManager will explode when using it in the postUp method of the first migration, because it will not run with entities matching this schema but with the latest version of entities.

https://github.com/doctrine/migrations/issues/124


stof:
thanks

PR #107, opened 2015-02-22
Allow symfony 3.0 components

nicolas-grekas:
Tests should tell if any deprecated interfaces of Symfony are used. If not, then the bundle is defacto compatible with 3.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     ],
     "require": {
         "php": ">=5.3.2",
-        "symfony/framework-bundle": "~2.1",
+        "symfony/framework-bundle": "~2.3|~3.0",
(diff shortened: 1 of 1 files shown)


odino:
Thanks man! Though this will remove compatibiliy with 2.1/2.2 -- what shall we do?

nicolas-grekas:
2.1 and 2.2 are not supported. 2.3 is the oldest version supported.
IMHO, the master branch of this bundle should not support obsolete versions

odino:
fair enough -- @stof can we go ahead? we have tags for previous versions so to me everything looks fine

odino:
going ahead!

PR #112, opened 2015-05-06
We finally got a stable tag!

garak:


odino:
thanks man! :)

craue:
Tag "1.0.1" created from the "2.1.x" branch? That's a bit confusing. :smirk:
Plus, this tag is not usable yet via Composer as it still doesn't appear on Packagist.

stof:
Well, the branch version and the tagged versions are indeed totally messed up (was already the case for 1.0). the issue is that we had a 2.0.x branch for Symfony 2.0 before we figured that bundle should not have their version in sync with Symfony. We should delete this 2.0.x version (renaming the branch to ``legacy-2.0`` for instance) and fix the branch alias for master

xabbuh:
@craue @stof see #114 for the branch alias fix

PR #118, opened 2015-05-31
Fix typo in documentation

zatikbalazs:
| Q             | A
| ------------- | ---
| Doc fix?      | yes
| New docs?     | no
| Applies to    | all
| Fixed tickets |

In the part "Generating Migrations Automatically" there is a mapping
information sample for creating a new User entity. On the "Annotations"
tab the "name" property has a length of "255".

Having those double quotes around 255 is incorrect because running
"app/console doctrine:migrations:diff" throws an error. It expects
"length" to be an integer and not a string.

I removed the double quotes to prevent this error.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -223,7 +223,7 @@ for Doctrine's ORM:
             protected $id;
 
             /**
-             * @ORM\Column(type="string", length="255")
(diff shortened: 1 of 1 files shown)


mikeSimonson:
@zatikbalazs Thanks

PR #119, opened 2015-06-11
Allowed DoctrineMigrationsBundle to work with Symfony apps using DBAL only

guilhermeblanco:
I also fixed a bug of accepting optional `--em` value (it should be required).

### DIFF
--- a/Command/Helper/DoctrineCommandHelper.php
+++ b/Command/Helper/DoctrineCommandHelper.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ * This file is part of the Doctrine MigrationsBundle
+ *
+ * The code was originally distributed inside the Symfony framework.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ * (c) Doctrine Project, Guilehrme Blanco <guilhermeblanco@hotmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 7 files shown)


mikeSimonson:
Thanks @guilhermeblanco 
Can you add tests ?

guilhermeblanco:
@mikeSimonson this bundle doesn't even have infrastructure for tests yet. That would be a task for another PR, not this one. =)

mikeSimonson:
@guilhermeblanco What do you mean the em should be required. I think it's only required for the Diff command.

guilhermeblanco:
@mikeSimonson sorry, I probably had to explain better.
Before it was allowed to run this:

```bash
php app/console doctrine:migrations:diff --em
```

Without requiring a value for `em`.

guilhermeblanco:
@ryancastle I already merged a fix for that from @xabbuh

Command/Helper/DoctrineCommandHelper.php
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Provides some helper and convenience methods to configure doctrine commands in the context of bundles
+ * and multiple connections/entity managers.
+ *
+ * @author Guilherme Blanco <guilhermeblanco@hotmail.com>
+ */
+abstract class DoctrineCommandHelper extends BaseDoctrineCommandHelper
+{
+    public static function setApplicationHelper(Application $application, InputInterface $input)
```
stof: the naming is weird. It is not setting an application helper

PR #120, opened 2015-06-16
don't call a method inside empty()

xabbuh:
Prior to PHP 5.5, `empty()` only handles variables.

### DIFF
--- a/Command/Helper/DoctrineCommandHelper.php
+++ b/Command/Helper/DoctrineCommandHelper.php
@@ -29,8 +29,9 @@ abstract class DoctrineCommandHelper extends BaseDoctrineCommandHelper
     public static function setApplicationHelper(Application $application, InputInterface $input)
     {
         $doctrine  = $application->getKernel()->getContainer()->get('doctrine');
+        $managerNames = $doctrine->getManagerNames();
 
-        if (empty($doctrine->getManagerNames())) {
+        if (empty($managerNames)) {
             self::setApplicationConnection($application, $input->getOption('db'));
         } else {
             self::setApplicationEntityManager($application, $input->getOption('em'));


ryancastle:
Is there anything wrong with:

    if ($doctrine->getManagerNames()) {
        self::setApplicationEntityManager($application, $input->getOption('em'));
    } else {
        self::setApplicationConnection($application, $input->getOption('db'));
    }

See http://stackoverflow.com/a/4328049/563394

alterphp:
+1

alterphp:
@ryancastle On my opinion, it should not be acceptable in a library/framework as it depends on the return type of Doctrine Registry::getManagerNames() method that comes from a third-party library.

xabbuh's patch solves the problem.

jroussel:
Hello and thank you for this quick fix.
I still have a problem with composer as it seems that the tag for version "2.1.x-dev" on packagist (https://packagist.org/packages/doctrine/doctrine-migrations-bundle#dev-master) is linked to previous commit (e33e07137c43a492589e89d79d36807cbfcf28e0)
Is it some kind of cache issue or could it be solved too?

stof:
@beberlei can you check whether the packagist webhook is configured properly ?

gabriperego:
Same issue here updating the Bundle, still getting the old commit.

smatyas:
We're still getting the old commit from packagist as well, had to fix the package for the previous commit hash. Please fix this issue, even by just triggering a manual update. Thanks in advance! :)

mikeSimonson:
@smatyas won't the --prefer-source option temporarily fix that ?

smatyas:
That doesn't work with our deploy scripts (yet), but thanks for the idea.

stof:
@mikeSimonson no it won't. ``--prefer-source`` will still checkout the commit referenced by Packagist. We need to trigger an update on Packagist (only @beberlei can do it as he is the only maintainer of the package there) and ensure we have the github hook configured (otherwise the package will be updated only once a week on Packagist)

Remiii:
Do you have some news about the Packagist update? Because on my side is still broken on PHP `5.3.x`.

smatyas:
FYI: It's in sync again (because of the weekly update, I assume).

PR #126, opened 2015-07-28
make the yaml comment syntax valid

greg0ire:
`//` works in php, but does not mean anything in yaml


greg0ire:
The code snippet is not colored, this was probably the cause.

xabbuh:
@greg0ire If you want to have it coloured, you will also have to enclose the "dir_name" value with quotes (due to the old YAML parser used by Sphinx).

greg0ire:
@xabbuh : thanks for the tip, done

stof:
thanks

PR #143, opened 2016-01-12
Adding a data collector for the debug toolbar

mikeSimonson:
The change to dev-master for doctrine/migrations in the composer.json is temporary.
As soon as the 1.3 version of the migrations is out it will target that.

### DIFF
--- a/Collector/MigrationsCollector.php
+++ b/Collector/MigrationsCollector.php
@@ -1,9 +1,14 @@
 <?php
-/**
- * Created by PhpStorm.
- * User: mike
- * Date: 9/01/16
- * Time: 23:02
+
+/*
+ * This file is part of the Doctrine MigrationsBundle
+ *
+ * The code was originally distributed inside the Symfony framework.
+ *
+ * (c) Michael Simonson <mike@simonson.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
  */
 
 namespace Doctrine\Bundle\MigrationsBundle\Collector;
--- a/Collector/MigrationsCollector.php
+++ b/Collector/MigrationsCollector.php
@@ -1,0 +1,165 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: mike
+ * Date: 9/01/16
+ * Time: 23:02
+ */
+
+namespace Doctrine\Bundle\MigrationsBundle\Collector;
+
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Migrations\Configuration\Configuration;
+use Doctrine\DBAL\Migrations\Tools\Console\Helper\MigrationStatusInfosHelper;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\DataCollector\DataCollectorInterface;
+
+class MigrationsCollector implements DataCollectorInterface
+{
+    /** @var  string */
+    private $migrationTablename;
+
+    /** @var  string */
+    private $migrationName;
+
+    /** @var Connection  */
+    private $connection;
+
+    /** @var  string */
+    private $migrationNamespace;
+
+    /** @var  string */
+    private $migrationDirectory;
+
+    public function __construct(Connection $connection, $migrationNamespace, $migrationDirectory, $migrationTablename, $migrationName)
+    {
+        $this->connection = $connection;
+        $this->migrationNamespace = $migrationNamespace;
+        $this->migrationDirectory = $migrationDirectory;
+        $this->migrationTablename = $migrationTablename;
+        $this->migrationName = $migrationName;
+    }
+
+    /**
+     * Collects data for the given Request and Response.
+     *
+     * @param Request $request A Request instance
+     * @param Response $response A Response instance
+     * @param \Exception $exception An Exception instance
+     */
+    public function collect(Request $request, Response $response, \Exception $exception = null)
+    {
+        $configuration = new Configuration($this->connection);
+        $configuration->setMigrationsNamespace($this->migrationNamespace);
+        $configuration->setMigrationsDirectory($this->migrationDirectory);
+        $configuration->setMigrationsTableName($this->migrationTablename);
+        $configuration->setName($this->migrationName);
+        $configuration->registerMigrationsFromDirectory($configuration->getMigrationsDirectory());
+        $migrationsStatusInfos = new MigrationStatusInfosHelper($configuration);
+        $this->data = $migrationsStatusInfos->getMigrationsInfos();
+        $newMigrationsList = $configuration->getMigrationsToExecute('up', $configuration->getLatestVersion());
+        $this->data['newMigrationsList'] = array_map(function($migration) {
+            return $migration->getVersion();
+        }, $newMigrationsList);
+        $this->data['executedUnavailableMigrationsList'] = array_map(function($migration) {
+            return $migration->getVersion();
+        }, $migrationsStatusInfos->getExecutedUnavailableMigrations());
+
+        $this->connection = null;
+        $this->migrationDirectory = null;
+        $this->migrationNamespace = null;
+        $this->migrationName = null;
+        $this->migrationTablename = null;
+    }
+
+    /**
+     * Returns the name of the collector.
+     *
+     * @return string The collector name
+     */
+    public function getName()
+    {
+        return 'doctrine.migrations_collector';
+    }
+
+    public function getPreviousMigration()
+    {
+        return $this->data['Previous Version'];
+    }
+
+    public function getCurrentMigration()
+    {
+        return $this->data['Current Version'];
+    }
+
+    public function getNextMigration()
+    {
+        return $this->data['Next Version'];
+    }
+
+    public function getLatestMigration()
+    {
+        return $this->data['Latest Version'];
+    }
+
+    public function getExecutedMigrations()
+    {
+        return $this->data['Executed Migrations'];
+    }
+
+    public function getExecutedUnavailableMigrations()
+    {
+        return $this->data['Executed Unavailable Migrations'];
+    }
+
+    public function getAvailableMigrations()
+    {
+        return $this->data['Available Migrations'];
+    }
+
+    public function getNewMigrations()
+    {
+        return $this->data['New Migrations'];
+    }
+
+    public function getNewMigrationsList()
+    {
+        return $this->data['newMigrationsList'];
+    }
+
+    public function getExecutedUnavailableMigrationsList()
+    {
+        return $this->data['executedUnavailableMigrationsList'];
+    }
+
+    public function getDatabaseDriver()
+    {
+        return $this->data['Database Driver'];
+    }
+    public function getDatabaseName()
+    {
+        return $this->data['Database Name'];
+    }
+    public function getConfigurationSource()
+    {
+        return $this->data['Configuration Source'];
+    }
+    public function getVersionTableName()
+    {
+        return $this->data['Version Table Name'];
+    }
+    public function getVersionColumnName()
+    {
+        return $this->data['Version Column Name'];
+    }
+    public function getMigrationNamespace()
+    {
+        return $this->data['Migrations Namespace'];
+    }
+    public function getMigrationDirectory()
+    {
+        return $this->data['Migrations Directory'];
+    }
+}
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -24,6 +24,18 @@ use Symfony\Component\Config\Definition\ConfigurationInterface;
  */
 class Configuration implements ConfigurationInterface
 {
+    private $debug;
+
+    /**
+     * Constructor
+     *
+     * @param Boolean $debug Whether to use the debug mode
+     */
+    public function __construct($debug)
+    {
+        $this->debug = (Boolean) $debug;
+    }
+
     /**
      * Generates the configuration tree.
      *
@@ -40,6 +52,7 @@ class Configuration implements ConfigurationInterface
                 ->scalarNode('namespace')->defaultValue('Application\Migrations')->cannotBeEmpty()->end()
                 ->scalarNode('table_name')->defaultValue('migration_versions')->cannotBeEmpty()->end()
                 ->scalarNode('name')->defaultValue('Application Migrations')->end()
+                ->booleanNode('profiler_enabled')->defaultValue($this->debug)->end()
             ->end()
         ;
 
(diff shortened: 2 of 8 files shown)


mikeSimonson:
@stof Any comments ?

mikeSimonson:
@stof Do you know how to conditionally enable that feature ? With a configuration so that it can be disabled by default ( for test ).

composer.json
```diff
@@ -23,7 +23,7 @@
         "php": ">=5.4.0",
         "symfony/framework-bundle": "~2.3|~3.0",
         "doctrine/doctrine-bundle": "~1.0",
-        "doctrine/migrations": "~1.0"
+        "doctrine/migrations": "dev-master"
```
stof: use `~1.3@dev`, not `dev-master`. This way, forgetting to change it won't hurt (as the stability flag is root-only anyway). This is the power of branch aliases

Resources/config/services.xml
```diff
@@ -0,0 +1,21 @@
...
+        <service id="doctrine_bundle.migrations_collector"
+                 class="Doctrine\Bundle\MigrationsBundle\Collector\MigrationsCollector"
+                 public="true"
+        >
+            <argument type="service" id="doctrine.dbal.default_connection"/>
+            <argument>%doctrine_migrations.namespace%</argument>
+            <argument>%doctrine_migrations.dir_name%</argument>
+            <argument>%doctrine_migrations.table_name%</argument>
+            <argument>%doctrine_migrations.name%</argument>
+            <tag name="data_collector" template="DoctrineMigrationsBundle::migrationsCollector.html.twig" id="doctrine_bundle.migrations_collector" priority="249"/>
```
stof: please use the `@DoctrineMigrations` notation instead, for consistency with other templates of the profiler

Resources/views/migrationsCollector.html.twig
```diff
@@ -0,0 +1,170 @@
...
+            {% set color = collector.executedUnavailableMigrations > 0 ? 'red' : '' %}
+            {% if color is empty %}
+                {% set color = 'yellow' %}
+            {% endif %}
+
+            <img width="20" height="28" src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABoAAAAcCAYAAAB/E6/TAAABLUlEQVR42u3TP0vDQBiA8UK/gDiLzi0IhU4OEunk5OQUAhGSOBUCzqWfIKSzX8DRySF0URCcMjWLIJjFD9Cpk/D6HITecEPUuzhIAz8CIdyTP/f2iqI4qaqqDx8l5Ic2uIeP/bquezCokOAFF+oCN3t4gPzSEjc4NEPaCldQbzjELTYW0RJzHDchwwem+ons6ZBpLSJ7nueJC22h0V+FzmwWV0ee59vQNV67CGVZJmEYbkNjfpY6X6I0Qo4/3RMmTdDDspuQVsJvgkP3IdMbIkIjLPBoadG2646iKJI0Ta2wxm6OdnP0/Tk6DYJgHcfxpw21RtscDTDDnaVZ26474GkkSRIrrPEv5sgMTfHe+cA2O6wPH6vOBpYQNALneHb96XTEDI6dzpEZ0VzO0Rf3pP5LMLI4tAAAAABJRU5ErkJggg==" alt="Doctrine Migrations">
+            <span class="sf-toolbar-info-piece-additional-detail">Migrations</span>
+            <span class="sf-toolbar-value sf-toolbar-status sf-toolbar-status-{{ color }}">{{ collector.executedMigrations }} / {{ collector.availableMigrations }}</span>
+        {% else %}
+        {% endif %}
```
stof: please support the new version of the profiler too, for Symfony 2.8+

Resources/views/migrationsCollector.html.twig
```diff
@@ -0,0 +1,170 @@
...
+    {% set color = collector.availableMigrations == collector.executedMigrations ? 'green' : '' %}
+    {% set color = collector.executedUnavailableMigrations > 0 ? 'red' : '' %}
+    {% if color is empty %}
+        {% set color = 'yellow' %}
+    {% endif %}
+
+    {# this is the content displayed as a panel in the toolbar #}
+    {% set icon %}
+        {% if profiler_markup_version == 1 %}
+            {% set color = collector.availableMigrations == collector.executedMigrations ? 'green' : '' %}
```
stof: this is useless, as you overwrite it on next line

Resources/config/services.xml
```diff
@@ -0,0 +1,21 @@
...
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services
+        http://symfony.com/schema/dic/services/services-1.0.xsd"
+>
+    <services>
+        <service id="doctrine_bundle.migrations_collector"
+                 class="Doctrine\Bundle\MigrationsBundle\Collector\MigrationsCollector"
+                 public="true"
+        >
+            <argument type="service" id="doctrine.dbal.default_connection"/>
```
stof: There is absolutely no guarantee that a connection named ``default`` exists.

And please inject the ManagerRegistry to keep the connection service lazy. This way, you will be able to avoid forcing the initialization of the connection just because the profiler

Resources/views/migrationsCollector.html.twig
```diff
@@ -0,0 +1,170 @@
...
+            {% set result = attribute(collector, 'get' ~ (property|capitalize) ) %}
+            <div class="sf-toolbar-info-piece">
+                <b>{{ property|replace({'Migrations': '', 'Migration': ''})|humanize }}</b>
+                <span>{{ result|replace({'<comment>': '', '</comment>': '', '(': '', ')': ''}) }}</span>
+            </div>
+        {% endfor %}
+    {% endset %}
+
+    {# the 'link' value set to 'false' means that this panel doesn't
+       show a section in the web profiler (default is 'true'). #}
```
stof: I suggest removing this comment

Resources/config/services.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services
+        http://symfony.com/schema/dic/services/services-1.0.xsd"
+>
+    <services>
+        <service id="doctrine_bundle.migrations_collector"
```
stof: Registering the collector should be configurable.

Resources/views/migrationsCollector.html.twig
stof: the content of the panel needs to support it too

Resources/views/migrationsCollector.html.twig
```diff
@@ -0,0 +1,170 @@
...
+            <span class="sf-toolbar-value sf-toolbar-status sf-toolbar-status-{{ color }}">{{ collector.executedMigrations }} / {{ collector.availableMigrations }}</span>
+        {% else %}
+        {% endif %}
+    {% endset %}
+
+    {% set text %}
+        {# this is the content displayed when hovering the mouse over
+           the toolbar panel #}
+        {% set properties = ['previousMigration', 'currentMigration', 'nextMigration', 'latestMigration', 'executedMigrations', 'executedUnavailableMigrations', 'availableMigrations', 'newMigrations'] %}
+        {% for property in properties%}
```
stof: missing space according to the Twig CS.

Collector/MigrationsCollector.php
```diff
@@ -0,0 +1,165 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: mike
+ * Date: 9/01/16
+ * Time: 23:02
+ */
```
stof, line 12: should be replaced by the actual header

stof:
@mikeSimonson put a bundle configuration (as done in DoctrineBundle for instance) and make it default to ``%kernel.debug%`` (but let people change it as they want)

Resources/config/services.xml
mikeSimonson: Do you have any documentation on how to make a service registration optional ?

Resources/config/services.xml
```diff
@@ -0,0 +1,21 @@
...
+                 class="Doctrine\Bundle\MigrationsBundle\Collector\MigrationsCollector"
+                 public="true"
```
mikeSimonson: @stof Should I make the service public ?

Resources/config/services.xml
mikeSimonson: @stof What notation are you referring to ?

Resources/config/services.xml
mikeSimonson: @stof Am I on the right track with debug property ?
I still couldn't find any doc on how not to load the collector based on that property.

mikeSimonson:
@stof What's left to fix ?

Resources/config/services.xml
stof: no. Data collectors don't need to be public, so it is best to let the ContainerBuilder inline the service definition whenever possible by making it private.
Btw, we generally don't write ``public="true"`` as this is the default behavior.

Collector/MigrationsCollector.php
```diff
@@ -0,0 +1,170 @@
+<?php
+
+/*
+ * This file is part of the Doctrine MigrationsBundle
+ *
+ * The code was originally distributed inside the Symfony framework.
+ *
+ * (c) Michael Simonson <mike@simonson.be>
```
stof: the copyright header should stay the same than in other files of the project (this is not tracking the author of the file)

Collector/MigrationsCollector.php
```diff
@@ -0,0 +1,170 @@
...
+        }, $newMigrationsList);
+        $this->data['executedUnavailableMigrationsList'] = array_map(function($migration) {
+            return $migration->getVersion();
+        }, $migrationsStatusInfos->getExecutedUnavailableMigrations());
+
+        $this->connection = null;
+        $this->migrationDirectory = null;
+        $this->migrationNamespace = null;
+        $this->migrationName = null;
+        $this->migrationTablename = null;
```
stof: I think resetting them here can cause issues when subrequests are used (as they also collect the info)

DependencyInjection/Configuration.php
```diff
@@ -40,6 +52,7 @@ public function getConfigTreeBuilder()
                 ->scalarNode('namespace')->defaultValue('Application\Migrations')->cannotBeEmpty()->end()
                 ->scalarNode('table_name')->defaultValue('migration_versions')->cannotBeEmpty()->end()
                 ->scalarNode('name')->defaultValue('Application Migrations')->end()
+                ->booleanNode('profiler_enabled')->defaultValue($this->debug)->end()
```
stof: ``enable_profiler`` or ``profiling`` would be more consistent with naming used in existing bundles (DoctrineBundle is using ``profiling``)

DependencyInjection/DoctrineMigrationsExtension.php
```diff
@@ -32,13 +34,18 @@ class DoctrineMigrationsExtension extends Extension
...
 
         $config = $this->processConfiguration($configuration, $configs);
 
         foreach ($config as $key => $value) {
             $container->setParameter($this->getAlias().'.'.$key, $value);
         }
+
+        if ($config['profiler_enabled']) {
+            $loader = new Loader\XmlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
+            $loader->load('services.xml');
```
stof: given that this file contains only services related to the profiling (and must stay this was as it is loaded only when profiling is enabled), please name it ``profiler.xml`` or something like that. Otherwise, contributors may add other unrelated services in it in the future (and not notice the mistake as they will develop in an environment with the profiling enabled)

Resources/views/Collector/icon.svg
```diff
@@ -0,0 +1,826 @@
+
+
+
+
+<!DOCTYPE html>
+<html lang="en" class=" is-copy-enabled">
```
stof: html does not make sense. This needs to be a SVG file

stof: It looks like this contains the HTML of a github page

Resources/views/Collector/migrations.html.twig
```diff
@@ -0,0 +1,175 @@
...
+
+{% block toolbar %}
+    {% set profiler_markup_version = profiler_markup_version|default(1) %}
+
+    {# this is the content displayed as a panel in the toolbar #}
+    {% set icon %}
+        {% if profiler_markup_version == 1 %}
+            {% set color = collector.executedUnavailableMigrations > 0 ? 'red' : '' %}
+            {% if color is empty %}
+                {% set color = 'yellow' %}
```
stof: instead of putting it here, just put it in the ternary operator above.

and why is it yellow in all cases ?

stof: btw, you don't handle colors in the profiler v2 for Symfony 2.8+, which looks inconsistent

Resources/views/Collector/migrations.html.twig
```diff
@@ -0,0 +1,175 @@
+{% extends app.request.isXmlHttpRequest ? '@WebProfiler/Profiler/ajax_layout.html.twig' : '@WebProfiler/Profiler/layout.html.twig' %}
+
+{% block toolbar %}
+    {% set profiler_markup_version = profiler_markup_version|default(1) %}
+
+    {# this is the content displayed as a panel in the toolbar #}
+    {% set icon %}
+        {% if profiler_markup_version == 1 %}
+            {% set color = collector.executedUnavailableMigrations > 0 ? 'red' : '' %}
```
stof: As you put it in red when there are some unavailable migration executed, you should also display this info in the details

Resources/views/Collector/migrations.html.twig
```diff
@@ -0,0 +1,175 @@
...
+                <span>
+                    {{ collector.executedMigrations }} / {{ collector.availableMigrations }}
+                </span>
+            </span>
+        </span>
+
+    {% else %}
+
+        <span class="label {{ collector.executedUnavailableMigrations > 0 ? 'label-status-error' }}">
+                <span class="icon">{{ include('@DoctrineMigrations/Collector/icon.svg') }}</span>
```
stof: the indentation looks messed up here

composer.json
```diff
@@ -23,7 +23,7 @@
         "php": ">=5.4.0",
         "symfony/framework-bundle": "~2.3|~3.0",
         "doctrine/doctrine-bundle": "~1.0",
-        "doctrine/migrations": "~1.0"
+        "doctrine/migrations": "~1.3@dev"
```
stof: ``@dev`` is useless

stof:
Could you post screenshots of the toolbar and of the profiler panel, both in a Symfony 2.7- and in a Symfony 2.8+ case ?

JHGitty:
@mikeSimonson ping, do you continue this PR?

mikeSimonson:
Well I don't have time to finish it now anyway so you can have a go at it if you want.

goetas:
Implemented in https://github.com/doctrine/DoctrineMigrationsBundle/pull/407

Issue #148, opened 2016-02-13
Manage table "migration_versions" via doctrine:schema:drop / create?


tackerm:
I just noticed the following behaviour, which confused me quite a bit. Don't know if this is a bug, by design or just a missing feature:

I'm setting up a new instance of an existing Symfony project. After deploying the code, I run 

`app/console doctrine:schema:create`

to create the database schema. I notice that the table `migration_versions` is not being generated. Now I check the status of my migrations and (obviously) see the following:

```
app/console doctrine:migrations:status

 == Configuration

    >> Name:                                               Recruitment Bundle Migrations
    >> Database Driver:                                    pdo_mysql
    >> Database Name:                                      demo_data
    >> Configuration Source:                               manually configured
    >> Version Table Name:                                 migration_versions
    >> Version Column Name:                                version
    >> Migrations Namespace:                               DemoBundle\Migrations
    >> Migrations Directory:                               /vagrant/demo/app/DoctrineMigrations
    >> Previous Version:                                   Already at first version
    >> Current Version:                                    0
    >> Next Version:                                       2015-07-14 14:00:33 (20150714140033)
    >> Latest Version:                                     2016-02-11 16:43:17 (20160211164317)
    >> Executed Migrations:                                0
    >> Executed Unavailable Migrations:                    0
    >> Available Migrations:                               82
    >> New Migrations:                                     82
```

So the MigrationsBundle now thinks that it has to run all 82 migrations. Because I am using migrations to update a database schema from "old" to "new", the migrations expect an old database schema, which is obviously not the same as the newly created schema. So in other words, all my migrations are broken for this specific instance.

I can of course populate the `migration_versions` table manually with the `doctrine:migrations:version` command (or maybe with a fixture), but wouldn't it make sense to create and populate this table automatically during `doctrine:schema:create`?

Likewise, when running 

`app/console doctrine:schema:drop --force`

I would expect the `migration_versions` table to be removed automatically instead of having to clean things up manually.

mikeSimonson:
The 2 tools have a very different goal.
The ```doctrine:migrations:create``` tool exist to create the schema in the state described by the code whereas the main goal of doctrine migrations is to migrate data. If you have no data to migrate and can just use ```app/console doctrine:schema:create```, you have no use for the migrations.

tackerm:
> The doctrine:migrations:create tool exist to create the schema in the state described by the code

[...]

PR #149, opened 2016-02-15
Add doc for "doctrine:migrations:version" command

tackerm:
Not a full documentation, just two basic examples.

Ref #148


mikeSimonson:
@tackerm thanks

PR #151, opened 2016-02-22
Add xml schema definition

aar0nr:
Useful for IDE autocompletion.


mikeSimonson:
Why are you copy pasting that file?

aar0nr:
"Copy pasting" from what source are you referring? The master branch does not contain an XSD file.

mikeSimonson:
@aar0nr Sorry I am totally misread that.

PR #145, opened 2016-01-13
Adding organize_migrations configuration

dextervip:
Just a pull request that allows the developer to use file organization feature from doctrine migrations by adding organize_migrations under doctrine_migrations. I am not sure if is that placed correctly but here it goes!

PR #153, opened 2016-03-09
Added support of the organize migrations configuration.

Koc:
Untested yet. Alternative way for #145. @mikeSimonson I've used similar way like doctrine-bundle uses for proxy config generation.


mikeSimonson:
@Koc Thanks

PR #157, opened 2016-04-09
Fix Symfony 3.0 console path

peterkokot:
console script has been moved since Symfony 3.x from `app/console` to `bin/console`

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -62,6 +62,12 @@ You can configure the path, namespace, table_name and name in your ``config.yml`
 Usage
 -----
 
+.. caution::
+
+    If your application is based on Symfony 3, replace ``php app/console`` by
+    ``php bin/console`` before executing any of the console commands included
+    in this article.
+
 All of the migrations functionality is contained in a few console commands:
 
 .. code-block:: bash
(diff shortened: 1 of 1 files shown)


mikeSimonson:
@peterkokot Hi, thanks for the update of the doc, but your change are only valid for symfony >= 3.0.
The 2.x is still supported.

scaytrase:
Since these are just the docs, I think it's more easier just to mention to replace `app/console` with `bin/console` if library user has *sf3 directory layout*.

The problem is not only that 2.x is supported, but the 2.x upgraded to 3.x will have old directory layout, so binding docs to the new layout is initially wrong

xabbuh:
FYI, the SensioGeneratorBundle displays a generic message on top of the relevant pages (see, for example, http://symfony.com/doc/master/bundles/SensioGeneratorBundle/commands/generate_bundle.html):

> If your application is based on Symfony 3, replace `php app/console` by `php bin/console` before executing any of the console commands included in this article.

peterkokot:
Hello, sorry for a bit late reply. I guess I've missed this a bit.

@mikeSimonson, @scaytrase and @xabbuh updated as suggested. Since Symfony 2.8 will be supported until November 2019, yes, this will be better fixed then instead probably. For now I've replaced this back to the `php app/bin` and added caution on top of the usage section.

Thanks.

mikeSimonson:
@peterkokot Thanks

PR #158, opened 2016-04-11
[UPDATE] Add return statements of parent class

Neirda24:
* Forward the result of parent classes


mikeSimonson:
@Neirda24 Thanks

PR #161, opened 2016-04-11
short return

lucchese:


mikeSimonson:
@lucchese I agree that this is shorter but it becomes completely unreadable.

Issue #173, opened 2016-06-02
Unable to use preconfigured migrations table name


aar0nr:
This line is causing me problems:

https://github.com/doctrine/DoctrineMigrationsBundle/blob/master/Command/DoctrineCommand.php#L62

In my application, the `Doctrine\DBAL\Migrations\Configuration\Configuration` object is already configured and set on the doctrine migration commands. However, this conditional uses a `||` operator instead of a `&&` operator, and even though the configuration object has a table name already configured, it still reads it from the container.

goetas:
Closing this as the issue is not clear what is the propose for this change,  the pull request got closed, the issue is almost 4 years old, and the 1.x branch is abandoned.

PR #174, opened 2016-06-02
Don't set table name if already set

aar0nr:
Ref: #173


mikeSimonson:
@aar0nr Can you add a test for that change ?

aar0nr:
Closing since I am no longer using this feature.

PR #190, opened 2017-04-22
Allow to use migration diff when using only DBAL

fesor:
Right now migration:diff command is only working with ORM. Some developers even use ORM only to make it possible to sync database schema, even if they don't need or don't use ORM features.

Another use case is that there is a need to use both dbal and orm for different parts of applications. So for example some part of the schema is not defined by ORM metadata. In that case developer could implement chained schema provider (or just decorate OrmSchemaProvider) to add additional schema definition. `doctrine/migration` (see issue #226) package provides a way to specify custom schema provider which then could be used to compare schemas. 

This PR provides changes to allow usage of this command using just DBAL. It also reduce code duplication for handling sharding.

The open questions is how to pass correct schema provider implementation into `MigrationsDiffDoctrineCommand` constructor, but this could be done in another PR. Right now I just registering this command as service and pass needed service as dependency. But maybe there might be a better way.

### DIFF
--- a/Command/MigrationsDiffDoctrineCommand.php
+++ b/Command/MigrationsDiffDoctrineCommand.php
@@ -14,10 +14,7 @@
 
 namespace Doctrine\Bundle\MigrationsBundle\Command;
 
-use Doctrine\Bundle\DoctrineBundle\Command\Proxy\DoctrineCommandHelper;
 use Doctrine\DBAL\Migrations\Tools\Console\Command\DiffCommand;
-use Doctrine\DBAL\Sharding\PoolingShardConnection;
-use LogicException;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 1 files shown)


mikeSimonson:
Thanks for the analysis @fesor, this is a BC break I believe so it will have to wait for the release of a new major version.

stof:
@mikeSimonson are you sure it is a BC break ?

mikeSimonson:
@stof No. But I thaught it was as it add a requirement for anyone that use that command in some code.
What do you think ?

stof:
which requirement is it adding ?

stof:
yeah, options are never required (unless you write your own code in the command to make it fail when the option is not passed).

fesor:
There are no any BC breaks introduced by this PR.

mikeSimonson:
@fesor Thanks

PR #191, opened 2017-05-03
Updating index.rst

Zuhayer:
Updating the references of app/console with bin/console.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -64,8 +64,8 @@ Usage
 
 .. caution::
 
-    If your application is based on Symfony 3, replace ``php app/console`` by
-    ``php bin/console`` before executing any of the console commands included
+    If your application is based on Symfony <2.5.*, replace ``php bin/console`` by
+    ``php app/console`` before executing any of the console commands included
     in this article.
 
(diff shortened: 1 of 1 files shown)


Resources/doc/index.rst
```diff
@@ -64,8 +64,8 @@ Usage
 
 .. caution::
 
-    If your application is based on Symfony 3, replace ``php app/console`` by
-    ``php bin/console`` before executing any of the console commands included
+    If your application is based on Symfony <2.5.*, replace ``php bin/console`` by
```
xabbuh: should be "replace [...] **with**"

stof: why 2.5 ? Any 2.x version of Symfony uses `app/console` by default.

Zuhayer: @stof The directory structure of Symfony was changed from version 2.5 (google search), that is why I chose 2.5.*. 

I have only worked in 2.1, 3.0, 3.1, so I am not sure if 2.6-8 used app/console or bin/console.

xabbuh: @Zuhayer The directory structure of the Symfony Standard Edition was changed with Symfony 3.

PR #192, opened 2017-05-03
Update index.rst

Nidhognit:
All informations from http://docs.doctrine-project.org/projects/doctrine-migrations/en/latest/reference/managing_migrations.html

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -183,6 +183,45 @@ table of your production database. Regardless of what server you're on, you
 can always safely run this command to execute only the migrations that haven't
 been run yet on *that* particular database.
 
+Also, you can to give it a test! First lets just do a ``dry-run`` to make sure it produces the SQL we expect:
+
+.. code-block:: bash
+
+    $ php app/console doctrine:migrations:migrate --dry-run
+    Executing dry run of migration up to 20100416130452 from 0
+
+  >> migrating 20100416130452
+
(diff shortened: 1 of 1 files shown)


Resources/doc/index.rst
```diff
@@ -183,6 +183,45 @@ table of your production database. Regardless of what server you're on, you
...
+
+If everything looks good, you can remove the ``–dry-run`` option and actually execute the migration:
+
+.. code-block:: bash
+
+    $ php app/console doctrine:migrations:migrate
+    WARNING! You are about to execute a database migration that could result in schema changes and data lost. Are you sure you wish to continue? (y/n)y
+    Migrating up to 20100416130452 from 0
+
+  >> migrating 20100416130452
```
stof: indentation is broken here (and on similar lines)

Resources/doc/index.rst
```diff
@@ -183,6 +183,45 @@ table of your production database. Regardless of what server you're on, you
 can always safely run this command to execute only the migrations that haven't
 been run yet on *that* particular database.
 
+Also, you can to give it a test! First lets just do a ``dry-run`` to make sure it produces the SQL we expect:
```
stof: `lets` is not valid.

Thus the doc on symfony.com never uses `we`, always `you`. And the doc of this bundle currently follows this rule

stof:
Rather than duplicating the doc from the library (and then having to maintain it twice), it would be better to add a link to the library documentation instead

PR #210, opened 2017-10-12
Add command registration to fix Symfony 3.4 auto-registration deprecation

Jean-Beru:
Commands auto-registration is deprecated in Symfony 3.4. This PR will avoid Auto-registration of the command "Doctrine\Bundle\DoctrineCacheBundle\Command\xxx" is deprecated since Symfony 3.4 and won't be supported in 4.0. Use PSR-4 based service discovery instead. exceptions.

### DIFF
--- a/DoctrineMigrationsBundle.php
+++ b/DoctrineMigrationsBundle.php
@@ -14,6 +14,7 @@
 
 namespace Doctrine\Bundle\MigrationsBundle;
 
+use Symfony\Component\Console\Application;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
 
 /**
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -1,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="doctrine_migrations.diff_command" class="Doctrine\Bundle\MigrationsBundle\Command\MigrationsDiffDoctrineCommand">
+            <tag name="console.command" />
+        </service>
+        <service id="doctrine_migrations.execute_command" class="Doctrine\Bundle\MigrationsBundle\Command\MigrationsExecuteDoctrineCommand">
+            <tag name="console.command" />
+        </service>
+        <service id="doctrine_migrations.generate_command" class="Doctrine\Bundle\MigrationsBundle\Command\MigrationsGenerateDoctrineCommand">
+            <tag name="console.command" />
+        </service>
+        <service id="doctrine_migrations.latest_command" class="Doctrine\Bundle\MigrationsBundle\Command\MigrationsLatestDoctrineCommand">
+            <tag name="console.command" />
+        </service>
+        <service id="doctrine_migrations.migrate_command" class="Doctrine\Bundle\MigrationsBundle\Command\MigrationsMigrateDoctrineCommand">
+            <tag name="console.command" />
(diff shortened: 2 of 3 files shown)


DoctrineMigrationsBundle.php
```diff
@@ -24,4 +25,10 @@
  */
 class DoctrineMigrationsBundle extends Bundle
 {
+    /**
+     * {@inheritDoc}
+     */
+    public function registerCommands(Application $application)
```
ogizanagi: This can be reverted actually. See https://github.com/doctrine/DoctrineFixturesBundle/pull/205#discussion_r146107456 and https://github.com/symfony/symfony/pull/24630#discussion_r145772853

Resources/config/services.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="doctrine_migrations.diff_command" class="Doctrine\Bundle\MigrationsBundle\Command\MigrationsDiffDoctrineCommand">
```
chalasr: `public="false"` for all commands

DoctrineMigrationsBundle.php
Jean-Beru: Indeed but it avoids useless checks. See https://github.com/doctrine/DoctrineCacheBundle/pull/120#issuecomment-335927975 and https://github.com/doctrine/DoctrineBundle/pull/715#discussion_r144030327
So what should we do ?

Resources/config/services.xml
Jean-Beru: It will break BC because services are auto-registered as public services, no ?

nicolas-grekas: it should work with pre3.4 SFs

chalasr: NVM, it doesn't work before 3.1 while this bundle supports 2.3+. Too bad

Jean-Beru: OK. Done!

chalasr: Should be reverted, my bad sorry.

Jean-Beru: No problem. Hope I will not have to revert it again: smile:

DoctrineMigrationsBundle.php
mikeSimonson: IMO if it works without that method should be removed without any concern for the useless check that will probably removed when the old feature is going to be phased out. Do thoses check have any measurable performance impact anyway ?

chalasr: This bundle supports outdated versions of symfony, starting from 2.3 when defining commands as services was not possible yet. This would break compatibility with such versions so either doctrine bumps symfony requirements to to a maintained version (not relevant if this is a bugfix) it must be reverted, same for DoctrineCacheBundle and DoctrineBundle.

mikeSimonson: Thsi bundle should stop supporting unsupported version of symfony

Jean-Beru: OK. We can keep compatibility for now. SF4 will solve this problem anyway. Reverted !

chalasr: @mikeSimonson see https://github.com/doctrine/DoctrineMigrationsBundle/pull/212

DoctrineMigrationsBundle.php
```diff
@@ -14,6 +14,7 @@
 
 namespace Doctrine\Bundle\MigrationsBundle;
 
+use Symfony\Component\Console\Application;
```
weaverryan: minor, but this addition isn't needed now

fabpot: fixed in 5137def

PR #213, opened 2017-10-31
Add method_exists to avoid fatal error

ruudvanderweijde:
Related error:
```
  [Symfony\Component\Debug\Exception\UndefinedMethodException]
  Attempted to call an undefined method named "getCustomTemplate" of class "Doctrine\DBAL\Migrations\Configuration\Configuration".
```

This commit can be reverted when https://github.com/doctrine/migrations v1.6 is released.

Relates to https://github.com/doctrine/DoctrineMigrationsBundle/issues/207
Relates to https://github.com/doctrine/DoctrineMigrationsBundle/compare/v1.2.1...v1.3.0#commitcomment-24024900

PR #216, opened 2017-10-31
Unstable Custom Template Breaking

themizzi:
The custom template configuration is used in version 1.3.0 of this bundle; however, this is not in the current stable release of doctrine/migrations. Should be removed from the 1.3 line until it is stable in the required doctrine/migrations version.


Soullivaneuh:
Fixes #217

lcobucci:
#213 already solves this in the correct way.

PR #218, opened 2017-11-10
Use PHPUnit\Framework\TestCase instead of PHPUnit_Framework_TestCase

gabriel-caruso:
I use the `PHPUnit\Framework\TestCase` notation instead of `PHPUnit_Framework_TestCase` while extending our TestCases. This will help us migrate to PHPUnit 6, that [no longer support snake case class names](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

### DIFF
--- a/Tests/Command/DoctrineCommandTest.php
+++ b/Tests/Command/DoctrineCommandTest.php
@@ -30,6 +30,7 @@ class DoctrineCommandTest extends TestCase
             'doctrine_migrations.name' => 'test',
             'doctrine_migrations.table_name' => 'test',
             'doctrine_migrations.organize_migrations' => Configuration::VERSIONS_ORGANIZATION_BY_YEAR,
+            'doctrine_migrations.custom_template' => null,
         )));
     }
 }
--- a/Tests/Command/DoctrineCommandTest.php
+++ b/Tests/Command/DoctrineCommandTest.php
@@ -6,8 +6,9 @@ use Doctrine\Bundle\MigrationsBundle\Command\DoctrineCommand;
 use Doctrine\DBAL\Migrations\Configuration\Configuration;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\ParameterBag\ParameterBag;
+use PHPUnit\Framework\TestCase;
 
-class DoctrineCommandTest extends \PHPUnit\Framework\TestCase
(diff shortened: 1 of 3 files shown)


mikeSimonson:
Can you fix the build failure for php 7 ?

alcaeus:
Build failures are due to an outdated PHPUnit version. The constraint should be bumped to allow PHPUnit 5.7 and 6.4 as well.

gabriel-caruso:
Gonna work on `PHPUnit` versions 5 and 6 support :construction:

gabriel-caruso:
@alcaeus If I just update to `PHPUnit 5 or 6`, the same test `Doctrine\Bundle\MigrationsBundle\Tests\DependencyInjection\DoctrineCommandTest::testConfigureMigrations` fails, if we don't require `doctrine/migrations` bellow version `1.6`, but this time I got an exception:
```php
Symfony\Component\DependencyInjection\Exception\ParameterNotFoundException:
You have requested a non-existent parameter "doctrine_migrations.custom_template". Did you mean one of these: "doctrine_migrations.namespace", "doctrine_migrations.table_name"?
```

What should we do?

alcaeus:
I'm looking at this failure now. It doesn't seem related to the test update but something is definitely off.

alcaeus:
Ah, looks like this is indeed a problem with the test itself. This diff fixes the failure:
```
--- a/Tests/Command/DoctrineCommandTest.php
+++ b/Tests/Command/DoctrineCommandTest.php
@@ -30,6 +30,7 @@ class DoctrineCommandTest extends TestCase
             'doctrine_migrations.name' => 'test',
             'doctrine_migrations.table_name' => 'test',
             'doctrine_migrations.organize_migrations' => Configuration::VERSIONS_ORGANIZATION_BY_YEAR,
+            'doctrine_migrations.custom_template' => null,
         )));
     }
 }
```

Heads up: the test doesn't perform an assertion and is thus marked as risky.

gabriel-caruso:
@alcaeus Done. As `DoctrineCommand::configureMigrations` doesn't return anything, and risky tests don't break `Travis CI`, I think we should be ready :clap:

alcaeus:
> risky tests don't break Travis CI

Interesting, I'd expect it to break, but that's a story for a different pull request. Thanks @gabriel-caruso!

PR #220, opened 2017-11-25
Require versions below doctrine/migrations 1.6

### LINKED ISSUE #219: PHP 7.x tests are faling
Discovered in #218, [all `PHP 7.x` tests](https://travis-ci.org/doctrine/DoctrineMigrationsBundle/builds/307170286) are falling because of an [`doctrine/migrations` method declare with PHP 7 return type `?string`](https://github.com/doctrine/migrations/blob/master/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php#L363). How can we solve it? :thinking:

gabriel-caruso:
This fixes #219 discovered in #218.

Require versions below `doctrine/migrations` [`1.6`](https://github.com/doctrine/migrations/releases/tag/v1.6.0) that [has some methods with `PHP 7` return type declaration](https://github.com/doctrine/migrations/pull/545).

I just need to require versions below `1.2`, because of [version `1.3` required `PHP 5.5`](https://github.com/doctrine/migrations/releases/tag/v1.3.0), and we have tests with `PHP 5.4`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "php": ">=5.4.0",
         "symfony/framework-bundle": "~2.7|~3.3|~4.0",
         "doctrine/doctrine-bundle": "~1.0",
-        "doctrine/migrations": "^1.1"
+        "doctrine/migrations": "~1.2.0|~1.5.0"
     },
     "require-dev": {
         "phpunit/phpunit": "^4.8.36"


alcaeus:
This is not an appropriate solution: composer dependency resolution should normally do the job and not install a version requiring 7.1 on 7.0.

alcaeus:
Closing as `Invalid`: the error is produced by an outdated PHPUnit version that can't generate mock objects compatible with nullable types. Updating the phpunit constraint in composer.json to `^4.8.36|^5.7|^6.4` should do the trick.

PR #222, opened 2018-01-29
Update configuration options documentation

martin-georgiev:
Add the default value for `custom_template` key and mark it as available since v1.2


mikeSimonson:
@martin-georgiev Thanks

PR #223, opened 2018-02-03
update config default namespace for SF4

pool0022:
Use App\Migration as default namespace

For recent versions of Symfony with Flex, 3.4 & 4, best practice is to always use "App" as a root namespace, no matter what the application

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -28,9 +28,9 @@ class Configuration implements ConfigurationInterface
         $rootNode
             ->children()
                 ->scalarNode('dir_name')->defaultValue('%kernel.root_dir%/DoctrineMigrations')->cannotBeEmpty()->end()
-                ->scalarNode('namespace')->defaultValue('Application\Migrations')->cannotBeEmpty()->end()
+                ->scalarNode('namespace')->defaultValue('App\Migrations')->cannotBeEmpty()->end()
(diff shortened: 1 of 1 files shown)


greg0ire:
Maintainers, I think this should be squash merged, and the title / body of the PR should be reused as a commit message / body.

DependencyInjection/Configuration.php
```diff
@@ -28,9 +28,9 @@ public function getConfigTreeBuilder()
         $rootNode
             ->children()
                 ->scalarNode('dir_name')->defaultValue('%kernel.root_dir%/DoctrineMigrations')->cannotBeEmpty()->end()
-                ->scalarNode('namespace')->defaultValue('Application\Migrations')->cannotBeEmpty()->end()
+                ->scalarNode('namespace')->defaultValue('App\Migrations')->cannotBeEmpty()->end()
                 ->scalarNode('table_name')->defaultValue('migration_versions')->cannotBeEmpty()->end()
-                ->scalarNode('name')->defaultValue('Application Migrations')->end()
+                ->scalarNode('name')->defaultValue('App Migrations')->end()
```
mikeSimonson: I don't really see the point of changing the name here ?

pgnd: Simply for some semblance of reader's consistency with SF4's usage of "App" -- and pointedly NOT using 'old' "Application".

Nbd either way -- it's more the namespace that's at issue.

PR #224, opened 2018-03-12
Update index.rst

cafferata:
There is no reason for the use ContainerInterface statement when you use the ContainerAwareTrait.


alcaeus:
> There is no reason for the use ContainerInterface statement when you use the ContainerAwareTrait.

Yes there is: https://github.com/doctrine/DoctrineMigrationsBundle/blob/master/Command/DoctrineCommand.php#L93

`ContainerAwareTrait` saves you from implementing the `setContainer` method while the interface tells consuming code that it actually provides a `setContainer` method.

PR #229, opened 2018-06-06
Move phpstan and doctrine/coding-standards to require-dev instead of require.

ccovey:
Noticed that phpstan and coding-standards were in require while I'm guessing these are only necessary in require-dev.


alcaeus:
Thanks Cody!

PR #235, opened 2018-09-05
Some updates in the docs for modern Symfony apps

javiereguiluz:
This morning we were discussing about migrations in Symfony's Slack ... and I realized that the main docs for this bundle still reference too much to old Symfony apps. If you agree, we could update that. Thanks!


jwage:
👍cool. Thx!

PR #241, opened 2018-11-03
Minor

ThomasLandauer:
I'm assuming that "Symfony >= 2.4" only applied to the trait, not to the Interface - is that right?


alcaeus:
> I'm assuming that "Symfony >= 2.4" only applied to the trait, not to the Interface - is that right?

That is correct. Thanks for updating this!

Issue #243, opened 2018-11-16
Remove documentation for "Container Aware Migrations"?


j4nr6n:
Injecting or using application services in your Migrations is discouraged by many in the community. Which makes total sense. But there is also a section in the documentation about [Container Aware Migrations](https://symfony.com/doc/master/bundles/DoctrineMigrationsBundle/index.html#container-aware-migrations)
The documented means of getting the container isn't enough. [Migrations aren't registered as services](https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/4.2/config/services.yaml#L18) by default, so constructor injection doesn't work and the call to `setContainer()` never happens.

Should the docs be updated to add bits about registering Migrations as services? Or should that particular part of the docs be removed? Or something else?

jwage:
I woukd support removing it.

PR #244, opened 2018-11-19
Remove container aware migrations documentation

j4nr6n:
It's not recommended to use services in migrations because they could change or be removed, which could leave your database in an unpredictable state.

In addition, the examples don't work without some extra configuration to register migrations as services, which is also, probably, not a good idea.

See #243 for more information.


alcaeus:
Build failure is unrelated. Thanks @j4nr6n!

PR #250, opened 2019-01-04
fix indentation in config tree builder

kaznovac:
first `->end()` belongs to `->validate()`'s `ExprBuilder`, and the second one is now inline with it's `ScalarNodeDefinition` creator `->scalarNode('organize_migrations')->defaultValue(false)`


jwage:
Thanks!

PR #252, opened 2019-01-09
Update doctrine/migrations dependency to 2.0 now that stable is released

jwage:


xabbuh:
Do we still need the `composer.lock` file (was it added for Scrutinizer?)?

PR #256, opened 2019-01-18
added .gitattributes

rvitaliy:
excluded useless files for production

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,12 @@
+* text=auto
+
+/Tests export-ignore
+.doctrine-project.json export-ignore
+.gitattributes export-ignore
+.gitignore export-ignore
+.scrutinizer.yml export-ignore
+.travis.yml export-ignore
+composer.lock export-ignore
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1,13 @@
+* text=auto
+
+/Tests export-ignore
+.doctrine-project.json export-ignore
+.gitattributes export-ignore
+.gitignore export-ignore
+.scrutinizer.yml export-ignore
+.travis.yml export-ignore
+Changelog.md export-ignore
```
alcaeus: I'd opt for leaving this along with the readme, what do you think?

.gitattributes
```diff
@@ -0,0 +1,13 @@
+* text=auto
```
alcaeus: I think this should be safe but haven't used it before. Opinions @doctrine/doctrinecore?

ostrolucky: yeah seems ok for this repo. it's all source code

alcaeus:
Thanks @rvitaliy!

PR #257, opened 2019-02-08
fixed doctrine version for `custom_template`

gondo:
parameter `custom_template` is only available since version `1.3` as you can see v `1.2.1` doesn't include this yet: https://github.com/doctrine/DoctrineMigrationsBundle/blob/v1.2.1/DependencyInjection/Configuration.php

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -52,7 +52,7 @@ application:
         name: 'Application Migrations'
         # available in version >= 1.2. Possible values: "BY_YEAR", "BY_YEAR_AND_MONTH", false
         organize_migrations: false
-        # available in version >= 1.2. Path to your custom migrations template
+        # available in version >= 1.3. Path to your custom migrations template
         custom_template: ~
         all_or_nothing: false
 


gondo:
problem is, that even this is wrong :/
DoctrineMigrationBundle `1.3` introduced config `custom_template` but it still depend on old version of DoctrineMigration `^1.1.` what does not use custom templates.
Only DoctrineMigration `1.7` finally use custom template.
However because of other dependencies i am stuck with DoctrineMigration `1.5.` os even though I have DoctrineMigrationBundle `1.3.2`, I am not able to use `custom_template`

alcaeus:
> DoctrineMigrationBundle `1.3` introduced config `custom_template` but it still depends on old version of DoctrineMigration `^1.1.` what does not use custom templates.

`^1.1` means that the bundle requires **at least** version 1.1.0, but allows anything less than 2.0.0. Since the logic was built backwards compatible (e.g. the `custom_template` argument is ignored when an older version of `doctrine/migrations` is installed), there's no reason to change anything. It simply means that the functionality won't be available when using Symfony 2.7 because of the requirements to the newer `doctrine/migrations` version.

Upgrade away from Symfony 2.7 if you want to use this: your change in #258 won't fix your problem as all that would happen is that it will downgrade you to an older bundle version without changing anything else.

PR #258, opened 2019-02-08
removed ~2.7 as SF version for custom_template

gondo:
for more info see my comment on PR #257


gondo:
this might break some existing old code

alcaeus:
> this might break some existing old code

Which is why we're not doing this in a patch release.

PR #265, opened 2019-04-30
Add schema provider setting

thewilkybarkid:
As mentioned in #190 there's no setting for the diff command's schema provider. This adds a `schema_provider` setting.

Refs https://github.com/libero/content-store/pull/8#discussion_r279656101.

### DIFF
--- a/DependencyInjection/DoctrineMigrationsExtension.php
+++ b/DependencyInjection/DoctrineMigrationsExtension.php
@@ -7,6 +7,7 @@ namespace Doctrine\Bundle\MigrationsBundle\DependencyInjection;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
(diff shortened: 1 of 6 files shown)


DependencyInjection/DoctrineMigrationsExtension.php
```diff
@@ -49,4 +52,17 @@ public function getNamespace() : string
...
+     * @param string[] $config
+     */
+    private function configureSchemaProvider(array $config, ContainerBuilder $container) : void
+    {
+        if (! $config['schema_provider']) {
+            return;
+        }
+
+        $diffCommand = $container->findDefinition('doctrine_migrations.diff_command');
+        $diffCommand->setArgument(0, new Reference($config['schema_provider']));
```
SenseException: A check if a valid `$config['schema_provider']` value exists should happen.

goetas:
This is possible in 3.x by using the services section in the new configuration format.
See https://github.com/doctrine/DoctrineMigrationsBundle/blob/master/Resources/doc/index.rst#configuration

PR #270, opened 2019-06-14
Support check_database_platform option

andrew-demb:
Since doctrine/migrations 2.1.0 (https://github.com/doctrine/migrations/pull/784) version there are new option `check_database_platform`.

I added support for this option.

My questions:
1) Should I specify in doc comment about `available since version`?
2) PR should be to master, or 2.0 branch? **master**
3) Should I update composer dependencies, or increase minimal `doctrine/migrations` version to `^2.1` to avoid `method_exists` check? **Yes**

### DIFF
--- a/Command/DoctrineCommand.php
+++ b/Command/DoctrineCommand.php
@@ -85,6 +85,7 @@ abstract class DoctrineCommand extends BaseCommand
         $configuration->setMigrationsColumnLength($container->getParameter('doctrine_migrations.column_length'));
         $configuration->setMigrationsExecutedAtColumnName($container->getParameter('doctrine_migrations.executed_at_column_name'));
         $configuration->setAllOrNothing($container->getParameter('doctrine_migrations.all_or_nothing'));
+        $configuration->setCheckDatabasePlatform($container->getParameter('doctrine_migrations.check_database_platform'));
 
         // Migrations is not register from configuration loader
         if (! ($configuration instanceof AbstractFileConfiguration)) {
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -50,6 +50,7 @@ class Configuration implements ConfigurationInterface
                 ->scalarNode('all_or_nothing')->defaultValue(false)->end()
                 ->scalarNode('name')->defaultValue('Application Migrations')->end()
                 ->scalarNode('custom_template')->defaultValue(null)->end()
+                ->scalarNode('check_database_platform')->defaultValue(true)->end()
                 ->scalarNode('organize_migrations')->defaultValue(false)
                     ->info('Organize migrations mode. Possible values are: "BY_YEAR", "BY_YEAR_AND_MONTH", false')
                     ->validate()
--- a/Resources/config/schema/doctrine_migrations-1.0.xsd
+++ b/Resources/config/schema/doctrine_migrations-1.0.xsd
@@ -15,6 +15,7 @@
             <xsd:attribute name="all_or_nothing" type="xsd:boolean" />
             <xsd:attribute name="name" type="xsd:string" />
             <xsd:attribute name="custom_template" type="xsd:string" />
+            <xsd:attribute name="check_database_platform" type="xsd:boolean" />
             <xsd:attribute name="organize-migrations">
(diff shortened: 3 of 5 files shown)


andrew-demb:
CI failed due to phpstan issue - unknown `setCheckDatabasePlatform` method.

Command/DoctrineCommand.php
```diff
@@ -99,6 +99,10 @@ public static function configureMigrations(ContainerInterface $container, Config
             $configuration->setCustomTemplate($container->getParameter('doctrine_migrations.custom_template'));
         }
 
+        if (method_exists($configuration, 'isDatabasePlatformChecked')) {
```
SenseException: Shouldn't this check for `setCheckDatabasePlatform()`?

andrew-demb: I don't know flow for similar issues, so I just follow the same style as above (custom template).

I can change check to `setCheckDatabasePlatform` if you prefer.

SenseException: The one above is also using the checked method name `getCustomTemplate()` in the if-block. There would still be a `empty($configuration->getCustomTemplate())` even if the method was available in all versions of `$configuration`.

If you check for an existing method `isDatabasePlatformChecked()`, shouldn't this method be called and checked for `true` or `false` in the if-block too? If `isDatabasePlatformChecked()` is not relevant to be called, you have to check if the method `setCheckDatabasePlatform()` exists.

andrew-demb: Changed to check `set-` method.

Command/DoctrineCommand.php
```diff
@@ -99,6 +99,10 @@ public static function configureMigrations(ContainerInterface $container, Config
             $configuration->setCustomTemplate($container->getParameter('doctrine_migrations.custom_template'));
         }
 
+        if (method_exists($configuration, 'setCheckDatabasePlatform')) {
```
greg0ire: Why not bump the constraint to 2.1.0 instead of using such conditionals?

andrew-demb: It was one of my questions in PR description.

greg0ire: Oh right, sorry! Well you know my opinion now.

greg0ire:
> PR should be to master, or 2.0 branch?

There is no BC-break, so 2.0 is good.

alcaeus:
Just for the record: as this is a new feature (and not a bugfix) it should go into the master branch for release in the next minor version. Also, bumping a dependency version in a patch release also shouldn’t happen. Please re-target the PR to the master branch, which also means that we can safely bump the library dependency version. Thanks!

greg0ire:
Looks like I'm not 100% familiar with the workflow right now :sweat_smile:

andrew-demb:
Rebased to master.

Command/DoctrineCommand.php
SenseException: You removed the `method_exists` for `setCheckDatabasePlatform()`. Did you find out that checking if this method exists isn't needed?

andrew-demb: It's not needed now, because I increase minimal `doctrine/migrations` version to 2.1

SenseException:
On a side-note: The configurations don't validate the input (e.g. if a config expectes a boolean value), but this should be done in a separate PR.

andrew-demb:
Rebased again to avoid travis failure (previously cs checks failed for unrelated files)

alcaeus:
@goetas with a new major coming up, I believe this should re-target the last 2.x release instead. Can I let you handle rebasing the PR when you revisit the bundle? Thanks!

goetas:
This has already been implemented in . @andrew-demb  are you interested in rebasing this for 2.x on top of the `2.3.x` branch?

andrew-demb:
Rebased to latest `2.1.x` (@alcaeus I can't find `2.3.x` branch, as you mention in https://github.com/doctrine/DoctrineMigrationsBundle/pull/270#issuecomment-576212358)

goetas:
Targeting 2.2 since this is a new feature

goetas:
Thanks @andrew-demb

PR #275, opened 2019-07-15
Lazy console commands

vudaltsov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

Since Symfony `3.4` console commands can be lazy thanks to [a static name property](https://github.com/symfony/symfony/pull/23887).
This change does not break BC because the BC layer is already [in the base command's constructor](https://github.com/symfony/symfony/pull/23887/files#diff-8e80595663798837db1b033187835535R75).

### DIFF
--- a/Command/MigrationsDiffDoctrineCommand.php
+++ b/Command/MigrationsDiffDoctrineCommand.php
@@ -19,12 +19,14 @@ use function assert;
  */
 class MigrationsDiffDoctrineCommand extends DiffCommand
 {
+    /** @var string */
+    protected static $defaultName = 'doctrine:migrations:diff';
+
     protected function configure() : void
     {
         parent::configure();
 
         $this
-            ->setName('doctrine:migrations:diff')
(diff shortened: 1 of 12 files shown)


vudaltsov:
Relates to https://github.com/doctrine/migrations/pull/839

stof:
This requires bumping the min version of `doctrine/migrations` to the version containing https://github.com/doctrine/migrations/pull/839, otherwise it would break BC (if the parent class still calls `setName` in the configure method, it will override the default name set here).

vudaltsov:
@stof, agree, thought about this at the breakfast today :)
So this PR should be on hold until doctrine/migrations#839 is released

alcaeus:
2.2.0 has been released, I've pushed a commit to require that version.

vudaltsov:
thank you, @alcaeus

alcaeus:
Thanks @vudaltsov!

PR #276, opened 2019-07-30
[DOC] Update migrations namespace

deguif:
Sometimes `App\Migrations` namespace is used, and sometimes it's the appropriate namespace that is used `DoctrineMigrations`.

This PR unifies the namespace in the documentation by always using `DoctrineMigrations`.
This was confusing me on the Symfony documentation website.


alcaeus:
Thanks @deguif!

PR #279, opened 2019-10-17
Allow Symfony 5 components

l-vo:
cc @nicolas-grekas

PR #280, opened 2019-10-20
Fix typo in dep name

nicolas-grekas:
Fixes the issue reported by @jwage in https://github.com/doctrine/DoctrineMigrationsBundle/pull/279#issuecomment-544259961

https://github.com/doctrine/migrations has the same issue

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^5.7|^6.4|^7.0",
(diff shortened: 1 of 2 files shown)


l-vo:
@nicolas-grekas thank you, strange that I had no problem locally and that packagist only reports the problem after this PR 😕

greg0ire:
Thanks @nicolas-grekas!

l-vo:
@stof Ok ! Thanks :)

PR #295, opened 2020-01-01
Symfony 5 parameter fix

michaljusiega:
Related to https://github.com/symfony/symfony/pull/28810

### DIFF
--- a/Command/DoctrineCommand.php
+++ b/Command/DoctrineCommand.php
@@ -45,7 +45,7 @@ abstract class DoctrineCommand extends BaseCommand
             $configuration->setMigrationsDirectory($dir);
         } else {
             // class Kernel has method getKernelParameters with some of the important path parameters
-            $pathPlaceholderArray = ['kernel.root_dir', 'kernel.cache_dir', 'kernel.logs_dir'];
+            $pathPlaceholderArray = ['kernel.project_dir', 'kernel.cache_dir', 'kernel.logs_dir'];
 
             foreach ($pathPlaceholderArray as $pathPlaceholder) {
                 if (! $container->hasParameter($pathPlaceholder) || ! preg_match('/\%' . $pathPlaceholder . '\%/', $dir)) {
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -41,7 +41,7 @@ class Configuration implements ConfigurationInterface
 
         $rootNode
             ->children()
-                ->scalarNode('dir_name')->defaultValue('%kernel.root_dir%/DoctrineMigrations')->cannotBeEmpty()->end()
+                ->scalarNode('dir_name')->defaultValue('%kernel.project_dir%/DoctrineMigrations')->cannotBeEmpty()->end()
                 ->scalarNode('namespace')->defaultValue('Application\Migrations')->cannotBeEmpty()->end()
                 ->scalarNode('table_name')->defaultValue('migration_versions')->cannotBeEmpty()->end()
                 ->scalarNode('column_name')->defaultValue('version')->end()
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -41,7 +41,7 @@ class Configuration implements ConfigurationInterface
 
         $rootNode
             ->children()
-                ->scalarNode('dir_name')->defaultValue('%kernel.project_dir%/DoctrineMigrations')->cannotBeEmpty()->end()
+                ->scalarNode('dir_name')->defaultValue('%doctrine.migrations.dir%')->cannotBeEmpty()->end()
                 ->scalarNode('namespace')->defaultValue('Application\Migrations')->cannotBeEmpty()->end()
                 ->scalarNode('table_name')->defaultValue('migration_versions')->cannotBeEmpty()->end()
                 ->scalarNode('column_name')->defaultValue('version')->end()
--- a/DependencyInjection/DoctrineMigrationsExtension.php
+++ b/DependencyInjection/DoctrineMigrationsExtension.php
@@ -23,6 +23,12 @@ class DoctrineMigrationsExtension extends Extension
     {
         $configuration = new Configuration();
 
+        if ($container->hasParameter('kernel.root_dir')) {
+            $container->setParameter('doctrine.migrations.dir', '%kernel.root_dir%/DoctrineMigrations');
+        } else {
+            $container->setParameter('doctrine.migrations.dir', '%kernel.project_dir%/src/Migrations');
+        }
+
         $config = $this->processConfiguration($configuration, $configs);
 
         foreach ($config as $key => $value) {
--- a/Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
+++ b/Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
@@ -35,7 +35,7 @@ class DoctrineMigrationsExtensionTest extends TestCase
             'kernel.bundles' => [],
             'kernel.cache_dir' => sys_get_temp_dir(),
             'kernel.environment' => 'test',
-            'kernel.root_dir' => __DIR__ . '/../../', // src dir
+            'kernel.project_dir' => __DIR__ . '/../../', // src dir
         ]));
     }
 }


greg0ire:
The parameter was introduced in https://github.com/symfony/symfony/pull/22315 , and available since 3.4.0 , so I think this should target 2.1.x instead of master.

michaljusiega:
Changed. I think I changed it well.

greg0ire:
Changing the base branch is the first step, and you did it correctly. The second step is to get rid of commits that are on master and which you did not author

1. Fetch all new commits: `git fetch --all`. ⬇️ 
2. Rebase on what you fetched interactively: ✂️ 
    1. `git rebase -i origin/2.1.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin repository.
    2. A window will show up with many lines, remove every line but the last one, which should correspond to your commit.
   2. If you run into a conflict: 💥 
        1. fix it with `git mergetool`. 😷 
        1. continue on your merry way with `git rebase --continue`. ⏩ 
3. Force push to overwrite all this : `git push --force`. ⬆️

michaljusiega:
It's too hard for me :< Sorry, I give up :/

greg0ire:
Don't worry, I will do it for you

greg0ire:
All is well now, although a bit confusing because you used your `master` branch for this :sweat_smile:

michaljusiega:
Thank you.

DependencyInjection/Configuration.php
```diff
@@ -41,7 +41,7 @@ public function getConfigTreeBuilder() : TreeBuilder
 
         $rootNode
             ->children()
-                ->scalarNode('dir_name')->defaultValue('%kernel.root_dir%/DoctrineMigrations')->cannotBeEmpty()->end()
+                ->scalarNode('dir_name')->defaultValue('%kernel.project_dir%/DoctrineMigrations')->cannotBeEmpty()->end()
```
xabbuh: In many cases, this now changes the default value if an application is based on the Symfony Standard Edition.

michaljusiega:
Can We finish this PR ?

DependencyInjection/Configuration.php
goetas: Isnt project dir and root dir the same? If not, than this should be considered a BC break.

SenseException: This old blog about a Symfony 3 and 4 version explains it more in detail: https://symfony.com/blog/new-in-symfony-3-3-a-simpler-way-to-get-the-project-root-directory

This seems like a BC break when DoctrineMigrationsBundle gets updated to a version with these changes.

michaljusiega: So, What should I do in this situation?

alcaeus: @xabbuh what's the best way to deal with this then? Should we move this change to the next major version and document it as a BC break?

dmaicher: What if we move the logic to php for deciding which parameter to use?

Like the following

```diff
diff --git a/DependencyInjection/Configuration.php b/DependencyInjection/Configuration.php
index c7adbdb..b5d3ff1 100644
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -41,7 +41,7 @@ class Configuration implements ConfigurationInterface
 
         $rootNode
             ->children()
-                ->scalarNode('dir_name')->defaultValue('%kernel.root_dir%/DoctrineMigrations')->cannotBeEmpty()->end()
+                ->scalarNode('dir_name')->defaultValue('%doctrine.migrations.dir%')->cannotBeEmpty()->end()
                 ->scalarNode('namespace')->defaultValue('Application\Migrations')->cannotBeEmpty()->end()
                 ->scalarNode('table_name')->defaultValue('migration_versions')->cannotBeEmpty()->end()
                 ->scalarNode('column_name')->defaultValue('version')->end()
diff --git a/DependencyInjection/DoctrineMigrationsExtension.php b/DependencyInjection/DoctrineMigrationsExtension.php
index 4a0e923..640c80b 100644
--- a/DependencyInjection/DoctrineMigrationsExtension.php
+++ b/DependencyInjection/DoctrineMigrationsExtension.php
@@ -23,6 +23,12 @@ class DoctrineMigrationsExtension extends Extension
     {
         $configuration = new Configuration();
 
+        if ($container->getParameterBag()->has('kernel.root_dir')) {
+            $container->setParameter('doctrine.migrations.dir', '%kernel.root_dir%/DoctrineMigrations');
+        } else {
+            $container->setParameter('doctrine.migrations.dir', '%kernel.project_dir%/src/Migrations');
+        }
+
         $config = $this->processConfiguration($configuration, $configs);
 
         foreach ($config as $key => $value) {
```

This would only change the default for Symfony 5+ apps where the parameter `%kernel.root_dir%` does not exist. Or am I missing something? :blush:

ro0NL: @dmaicher that produces the same default across all Symfony version, thus preserves BC :+1: 

related to https://github.com/symfony/recipes/issues/718 it could be considerable to deprecate _not_ setting `dir_name` (which should effect a small group, as the recipe already sets it since forever in most projects AFAIK).

That would allow doctrine to update the default in its next major; the recipe is proposing `%kernel.project_dir%/migrations`. So if we can keep the default in sync, we can upgrade the recipe (remove the config) once we depend on doctrine's next major.

alcaeus: We're making more significant changes in the next major, where there may be multiple paths for migrations, so I'm not too concerned about the upcoming BC break.

@dmaicher great idea with the change, that should work since we can check what version we're on and select the appropriate parameter. @michaljusiega can you make those changes please? The patch above should suffice to ensure BC.

michaljusiega: @alcaeus done, waiting for green CI

DependencyInjection/Configuration.php
```diff
@@ -41,7 +41,7 @@ public function getConfigTreeBuilder() : TreeBuilder
 
         $rootNode
             ->children()
-                ->scalarNode('dir_name')->defaultValue('%kernel.root_dir%/DoctrineMigrations')->cannotBeEmpty()->end()
+                ->scalarNode('dir_name')->defaultValue('%doctrine.migrations.dir%')->cannotBeEmpty()->end()
```
dmaicher: to stick with the naming of the other parameters I would name this one `%doctrine_migrations.default_dir_name%`

Command/DoctrineCommand.php
```diff
@@ -45,7 +45,7 @@ public static function configureMigrations(ContainerInterface $container, Config
             $configuration->setMigrationsDirectory($dir);
         } else {
             // class Kernel has method getKernelParameters with some of the important path parameters
-            $pathPlaceholderArray = ['kernel.root_dir', 'kernel.cache_dir', 'kernel.logs_dir'];
+            $pathPlaceholderArray = ['kernel.project_dir', 'kernel.cache_dir', 'kernel.logs_dir'];
```
dmaicher: I think this can be a BC break? The code below seems to replace/resolve parameter placeholders with kernel parameters. So now this will not work anymore if my Symfony 3.4 project relies on `kernel.root_dir`?

dmaicher: So I think for now we can just add `kernel.project_dir` to the array but also keep `kernel.root_dir`. There is a check below that ensures the kernel parameter exists.

DependencyInjection/DoctrineMigrationsExtension.php
```diff
@@ -23,6 +23,12 @@ public function load(array $configs, ContainerBuilder $container) : void
     {
         $configuration = new Configuration();
 
+        if ($container->hasParameter('kernel.root_dir')) {
+            $container->setParameter('doctrine.migrations.dir', '%kernel.root_dir%/DoctrineMigrations');
+        } else {
+            $container->setParameter('doctrine.migrations.dir', '%kernel.project_dir%/src/Migrations');
```
xabbuh: Sorry for the late response.

As far as I can see this would still be a BC break for someone having an application that uses the old directory layout (i.e. they only updated their dependencies, but didn't adapt the directory structure established with Symfony Flex).

I think the only safe upgrade path is to deprecate not configuring this option and making it a mandatory one in the next major release.

dmaicher: Indeed @xabbuh is right :cry: This will break for people with the old directory structure that just move to Symfony 5 without adapting to the new structure.

alcaeus: > I think the only safe upgrade path is to deprecate not configuring this option and making it a mandatory one in the next major release.

I'll defer this to @goetas, since the next major release will change the config format anyways. We may just want to ignore this for now and fix it properly in 3.0.

michaljusiega:
So what about this PR ? I can't bump to SF5 due to kernel.root_dir which does not exist.

PR #298, opened 2020-02-21
Detect the entity manager in a compiler pass

goetas:
Fixes https://github.com/doctrine/DoctrineMigrationsBundle/issues/297

This PR moves in a compiler pass the detection of the entity manager or dbal connection.

### DIFF
--- a/Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
+++ b/Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\Bundle\MigrationsBundle\Tests\DependencyInjection;
 
 use Doctrine\Bundle\MigrationsBundle\DependencyInjection\DoctrineMigrationsExtension;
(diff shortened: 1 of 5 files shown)


Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
```diff
@@ -280,43 +240,56 @@ public function testInvalidService() : void
...
             'connection' => 'custom',
         ];
 
+        $container = $this->getContainer($config);
+
+        $container->compile();
+    }
+
+    /**
+     * @param mixed $config
```
SenseException: This is supposed to be array.

Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
```diff
@@ -280,43 +240,56 @@ public function testInvalidService() : void
...
+     */
+    public function getContainer($config) : ContainerBuilder
```
SenseException: Does this need to be `public`?

goetas:
@SenseException thanks for the review! both things have been fixed

PR #299, opened 2020-03-02
[Bug] Drop usage of application name

chr-hertel:
see https://github.com/doctrine/migrations/commit/a096d2c34ea544d569e4edbc54452d7b6125a983#diff-2f01203d5125b167941bdd7f6b8e2fb5


goetas:
Thanks

PR #303, opened 2020-03-17
Remove unknown 2.2.x branch from config

SenseException:
The website build is breaking because it cannot fetch the branch 2.2.x from this repository. Either we delete the branch from the config, which is the content of this PR, or we create this missing branch for this repository.


alcaeus:
Thanks @SenseException!

PR #306, opened 2020-03-31
Update readme

riccardonar:
Removed entry `name` and entry duplicated `check_database_platform` in configuration instruction


goetas:
@senseexception you are right, should be removed from the configs as  well

PR #307, opened 2020-03-31
Update default table name

dmaicher:
The default name seems to be `doctrine_migration_versions`:

https://github.com/doctrine/migrations/blob/master/lib/Doctrine/Migrations/Metadata/Storage/TableMetadataStorageConfiguration.php#L10


goetas:
Thanks

PR #308, opened 2020-03-31
Add new doctrine:migrations:list command to docs

dmaicher:
I guess we forgot to add the new command there as well? :blush:


goetas:
Thanks

PR #317, opened 2020-04-24
Add new doctrine current command

goetas:
Integrates https://github.com/doctrine/migrations/pull/965 into the bundle

### DIFF
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -52,6 +52,14 @@
             <tag name="console.command" command="doctrine:migrations:list" />
         </service>
 
+        <service id="doctrine_migrations.current_command" class="Doctrine\Migrations\Tools\Console\Command\CurrentCommand">
+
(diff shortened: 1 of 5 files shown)


greg0ire:
Shouldn't the `doctrine/migrations` requirement be bumped?

goetas:
Currently we are requiting `doctrine/migrations:^3.0@dev`, until 3.0.0 is not released, no version bump is required

goetas:
But only now i see the test failure... but i have no idea how to fix it...

greg0ire:
Besides pinning to a given commit, I have no idea either

greg0ire:
Ah wait I know, try to use `master`. Right now it's installing the alpha1 tag, which matches `@dev`

goetas:
like this?

greg0ire:
yeah I think that should work

goetas:
yeah

PR #323, opened 2020-05-09
Lazy services and service factories

goetas:
- Services passed to the `DependencyFactory` are always lazy
- Services passed to the `DependencyFactory` can be build via callables, allowing to fetch other dependencies from the  `DependencyFactory`  itself

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -88,6 +88,21 @@ application:
             # Custom migration classes factory
             'Doctrine\Migrations\Version\MigrationFactory': ~
 
+        factories:
+            # Custom migration sorting service id via callables (MyCallableFactory must be a callable)
+            'Doctrine\Migrations\Version\Comparator': 'MyCallableFactory'
+
+
+
+
+- The ``services`` node allows you to provide custom services to the underlying ``DependencyFactory`` part
+of ``doctrine/migrations``.
+
+- The node ``factories`` is similar to ``services``, with the difference that it accepts only callables.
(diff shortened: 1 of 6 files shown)


Resources/doc/index.rst
```diff
@@ -88,6 +88,21 @@ application:
...
+            # Custom migration sorting service id via callables (MyCallableFactory must be a callable)
+            'Doctrine\Migrations\Version\Comparator': 'MyCallableFactory'
+
+
+
+
+- The ``services`` node allows you to provide custom services to the underlying ``DependencyFactory`` part
+of ``doctrine/migrations``.
+
+- The node  ``factories`` is similar to ``services``, with the difference that accepts only callables.
```
greg0ire, line 101:
```suggestion
- The node ``factories`` is similar to ``services``, with the difference that it accepts only callables.
```

Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
```diff
@@ -175,6 +177,69 @@ public function compare(Version $a, Version $b) : int
...
+    {
+        $config    = [
+            'services' => [Comparator::class => 'my_sorter'],
+        ];
+        $container = $this->getContainer($config);
+
+        $conn = $this->createMock(Connection::class);
+        $container->set('doctrine.dbal.default_connection', $conn);
+
+        $sorterFactory = new class(){
```
greg0ire, line 190:
```suggestion
        $sorterFactory = new class() {
```

PR #334, opened 2020-06-16
Update UPGRADE.md

### LINKED ISSUE #333: Upgrade From 2.x to 3.0.0 missing docs about using `custom` template
I have been used some custom template for migrations and seems like there is change that is missing from UPGRADE.md

Earlier:
`final class Version<version> extends AbstractMigration`
With 3.x:
`final class <className> extends AbstractMigration`

tarlepp:
Fixes #333

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -55,3 +55,17 @@ doctrine_migrations:
 After
 
 The parameter `name` has been dropped.
+
+- Custom migration templates:
+
+Before
+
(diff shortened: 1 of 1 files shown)


greg0ire:
I think you should target 3.0.x

tarlepp:
> I think you should target 3.0.x

true, thanks!

goetas:
Thanks !

PR #339, opened 2020-06-17
Merge 3.0.x into master

greg0ire:


greg0ire:
Scrutinizer is still here… I just removed the associated webhook, this should be the last we see of it.

greg0ire:
There, no more Scrutinizer :)

PR #340, opened 2020-06-19
Document table name BC break

goetas:
In `doctrine/migrations` 3.x and `doctrine/DoctrineMigrationsBundle` 3.x the default value for  `table name` is `doctrine_migration_versions`.

In the previous 2.x versions the doctrine/migrations library was using as default `doctrine_migration_versions` while the bundle was using `migration_versions`.

Closes https://github.com/doctrine/migrations/issues/1002

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -40,7 +40,7 @@ doctrine_migrations:
         table_storage:
             table_name: 'migration_versions'
             version_column_name: 'version'
-            version_column_length: 1024
+            version_column_length: 191
             executed_at_column_name: 'executed_at'
 ```
 - The migration name has been dropped:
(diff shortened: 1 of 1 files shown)


UPGRADE.md
```diff
@@ -56,6 +56,17 @@ After
 
 The parameter `name` has been dropped.
 
+
+- The default for `table_name` changed from `migration_versions` to `doctrine_migration_versions`. If you did not
+specify the `table_name` option, now  need to declare it explicitly to not loose migration data.
```
greg0ire, line 61:
```suggestion
specify the `table_name` option, you now need to declare it explicitly to not loose migration data.
```

guilliamxavier: also `loose` -> `lose`

goetas:
I have also removed the some upgrade instructions that are relative to the `doctrine/migrations` library and added a reference to its dedicated upgrading document.

PR #348, opened 2020-06-26
Document that container aware migrations are not special anymore

goetas:
closes https://github.com/doctrine/DoctrineMigrationsBundle/issues/349
closes https://github.com/doctrine/DoctrineMigrationsBundle/issues/350

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -72,3 +72,6 @@ doctrine_migrations:
 Upgrading this bundle to `3.0` will also update the `doctrine/migrations` library to the version `3.0`.
 Backward incompatible changes in `doctrine/migrations` 3.0 
 are documented in the dedicated [UPGRADE](https://github.com/doctrine/migrations/blob/3.0.x/UPGRADE.md) document. 
+
+- The container is not automatically injected anymore when a migration implements `ContainerAwareInterface`. Custom
+migration factories should be used to inject additional dependencies into migrations.
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -229,6 +229,67 @@ You can skip single migrations by explicitly adding them to the ``migration_vers
 
 Doctrine will then assume that this migration has already been run and will ignore it.
 
+Migration Dependencies
+----------------------
+
+Migrations can have dependencies on external services (as geolocation, mailer, data processing services...) that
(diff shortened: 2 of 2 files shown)


UPGRADE.md
```diff
@@ -69,3 +69,6 @@ After
 ```php
 final class <className> extends AbstractMigration
 ```
+
+- The container is not automatically injected anymore when a migration implements `ContainerAwareInterface`. Custom
+migration factories should to be used to inject additional dependencies into migrations.
```
greg0ire, line 74:
```suggestion
migration factories should be used to inject additional dependencies into migrations.
```

SenseException:
This information might be relevant for the documentation too. I don't know in which chapter it would fit in though.

goetas:
@SenseException i could not find any reference to this feature in the documentation, I think that was something available but never documented.

SenseException (approved):

rnbwspctrm:
I have spent days trying to figure out how to access the entity manager in v3.0 after upgrading from 1.3. The link on the read me on the master branch to "read the documentation for this bundle" links to the 1.3 version on the symfony docs. Even switching to the 3.0 docs manually does not reference these changes at all, nor any indication of how to access the em. 

How long does it take for the docs to update with important merged information? Is there some other standard way to modify large amounts of data in Symfony without using the entity manager in a migration? This is a very core use of this bundle for our project. Unless I'm missing something, (please share if I am!), adding a dedicated section about how to access the entity manager in the migrations is very important information to be documented.

fabpot:
Is there anywhere I can read about the decision to remove support for injecting the Container?

goetas:
@fabpot 
> Is there anywhere I can read about the decision to remove support for injecting the Container?

Not really. it has been done as part of the major refactoring done in v3. The rationale behind them is to remove the magic (some command injecting dependencies based on marker interfaces) and let people use dependency injection do the job by providing dependencies they might need.

goetas:
I have in plan to work on this and to provide a factory (enabled by default when using the bundle)that will provide the container if the migration is "container aware"

goetas:
https://github.com/doctrine/DoctrineMigrationsBundle/pull/368 will introduce a default migration factory that will check if the migration class implements the `ContainerAwareInterface` class.

PR #358, opened 2020-07-28
Upgrade instructions for projects without an existing metadata table definition.

bartfeenstra:
I spent some time figuring out why I couldn't get the upgrade to work, until I realized that the metadata table instructions seemed as if they weren't applicable if projects didn't originally override this configuration in the first place. This PR aims to introduce the minimal instructions needed to upgrade such projects.

Running the `doctrine:migrations:sync-metadata-storage` command results in the `execution_time` column being added with `NULL` values. I presume that's alright as it's done by default and running migrations appears to work successfully afterwards, but please do verify that.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -43,6 +43,15 @@ doctrine_migrations:
             version_column_length: 191
             executed_at_column_name: 'executed_at'
 ```
+If your project did not originally specify its own table definition configuration, you will need to configure the table name after the upgrade:
+
+```yaml
+doctrine_migrations:
+    storage:
+        table_storage:
+            table_name: 'migration_versions'
+```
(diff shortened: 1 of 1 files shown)


greg0ire:
Hi! Thanks for your contribution! Why did you target master? Any specific reason?

bartfeenstra:
Thanks for the reviews!

@greg0ire That was the default, and it appears to be the most up-to-date branch. Should I have targeted `3.0.x` instead?

greg0ire:
I changed the default branch 6 days ago, that's why I'm asking, and yeah, you should have targeted `3.0.x`, I rebased it for you right now :slightly_smiling_face:

goetas:
thank you all!

Issue #38, opened 2012-07-04
Multiple entity managers support


dinamic:
Inside our ERP system we have calls to multiple database storages. To use them with Symfony2, we've defined them as different connections each one attached to a different entity manager.

If we make some change to the database structure there is no way to specify on which entity manager this change is going to be performed.

This can be controlled via the command line, but doing so is not reliable, unless we cherry-pick the doctrine migrations on deployment and specify different "--em" argument depending on the current migration.

richsage:
+1

zodeus:
+1

hasumedic:
+1

suncat2000:
Awesome!

hason:
+1

aclarkd:
+1

odino:
+1

kingcrunch:
:+1: 

But seriously: What happened to this ticket?

dinamic:
Two years later nothing has changed.

I am still going to recommend the [Container Aware Migrations](https://github.com/doctrine/DoctrineMigrationsBundle/blob/master/Resources/doc/index.rst#container-aware-migrations) as a workaround.

Max101:
:thumbsup: +1 
Hope to see this someday...

ryall:
+1

pasmaill:
+1

eggyal:
+1

sp-enric-fradera:
A few downsides I found to the proposed container workaround:
- It's ugly as hell.
- You get a console warning for not executing any SQL.
- The queries get executed even if you do a dry-run.

spinx:
@stof 
We have a fork where we added support for this. Kinda necessary with 16 EMs on the app. 

Couple of things:
- we need to track the em migration was executed against like so http://i.imgur.com/aAhHfuh.png
- changes to doctrine/migrations are necessary

If we clean the code up a bit would you be in a position to merge a PR in on both this repo and doctrine/migrations ?

stof:
@spinx does it really need a change in doctrine/migrations itself ? what kind of change ?

spinx:
@stof Mostly changes to commands. To allow diffing on all EMs, generating migrations for EMs. As I mentioned the only downside is that you need to track which EM the migration is for so you need to append EM name or similar to the migration version.

mikeSimonson:
@spinx or you just use different configuration for different em and your problem is already solved.

spinx:
[...]

PR #364, opened 2020-09-11
Add manager loader support introduced in migrations 3.1 (dev)

goetas:
bundle support for https://github.com/doctrine/migrations/pull/1023

Closes https://github.com/doctrine/migrations/issues/1062

### DIFF
--- a/Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
+++ b/Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
@@ -17,6 +17,7 @@ use Doctrine\Migrations\Metadata\Storage\TableMetadataStorageConfiguration;
 use Doctrine\Migrations\Version\Comparator;
 use Doctrine\Migrations\Version\Version;
 use Doctrine\ORM\EntityManager;
+use Doctrine\Persistence\ManagerRegistry;
 use Exception;
 use InvalidArgumentException;
 use PHPUnit\Framework\TestCase;
@@ -297,6 +298,68 @@ class DoctrineMigrationsExtensionTest extends TestCase
         $di = $container->get('doctrine.migrations.dependency_factory');
 
         self::assertInstanceOf(DependencyFactory::class, $di);
+        self::assertSame($em, $di->getEntityManager());
+    }
+
+    /**
+     * @return array<array{string|null}>
+     */
+    public function getEmNames(): array
+    {
+        return [
+            [null],
+            ['custom'],
+        ];
+    }
+
+    /**
+     * @dataProvider getEmNames
+     */
+    public function testPrefersManagerRegistry(?string $emName): void
+    {
+        $config    = ['em' => $emName];
+        $container = $this->getContainer($config);
+
+        $em           = $this->createMock(EntityManager::class);
+        $registryMock = $this->createMock(ManagerRegistry::class);
+
+        $registryMock->expects(self::once())
+            ->method('getManager')
+            ->with($emName)
+            ->willReturn($em);
+
+        $container->set('doctrine', $registryMock);
+
+        $container->compile();
+
+        $di = $container->get('doctrine.migrations.dependency_factory');
+
+        self::assertInstanceOf(DependencyFactory::class, $di);
+        self::assertSame($em, $di->getEntityManager());
+    }
+
+    public function testManagerRegistryThrowsExceptionOnUnknownEm(): void
+    {
+        $config    = ['em' => 'foo'];
+        $container = $this->getContainer($config);
+
+        $em           = $this->createMock(EntityManager::class);
+        $registryMock = $this->createMock(ManagerRegistry::class);
+
+        $registryMock->expects(self::once())
+            ->method('getManager')
+            ->with('foo')
+            ->willThrowException(new InvalidArgumentException());
+
+        $container->set('doctrine', $registryMock);
+
+        $container->compile();
+
+        $di = $container->get('doctrine.migrations.dependency_factory');
+
+        self::assertInstanceOf(DependencyFactory::class, $di);
+
+        $this->expectException(InvalidArgumentException::class);
         self::assertSame($em, $di->getEntityManager());
     }
 
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "php": "^7.2|^8.0",
         "symfony/framework-bundle": "~3.4|~4.0|~5.0",
         "doctrine/doctrine-bundle": "~1.0|~2.0",
-        "doctrine/migrations": "^3.0.3"
+        "doctrine/migrations": "^3.1"
     },
     "require-dev": {
         "phpunit/phpunit": "^7.0|^8.0|^9.0",
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,8 @@
         "phpstan/phpstan-deprecation-rules": "^0.12",
         "phpstan/phpstan-phpunit": "^0.12",
         "phpstan/phpstan-strict-rules": "^0.12",
-        "doctrine/orm": "^2.6"
+        "doctrine/orm": "^2.6",
+        "doctrine/persistence": "^1.3||^2.0"
     },
     "autoload": {
         "psr-4": { "Doctrine\\Bundle\\MigrationsBundle\\": "" },
--- a/DependencyInjection/CompilerPass/ConfigureDependencyFactoryPass.php
+++ b/DependencyInjection/CompilerPass/ConfigureDependencyFactoryPass.php
@@ -17,27 +17,41 @@ class ConfigureDependencyFactoryPass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container): void
     {
-        $preferredEm  = $container->getParameter('doctrine.migrations.preferred_em');
         $diDefinition = $container->getDefinition('doctrine.migrations.dependency_factory');
+
+        $preferredEm = $container->getParameter('doctrine.migrations.preferred_em');
+        if ($container->has('doctrine')) {
+            $loaderDefinition = $container->getDefinition('doctrine.migrations.registry_loader');
+            $loaderDefinition->setArgument(0, new Reference('doctrine'));
+            if ($preferredEm !== null) {
+                $loaderDefinition->setArgument(1, $preferredEm);
+            }
+
+            $diDefinition->setFactory([DependencyFactory::class, 'fromEntityManager']);
+            $diDefinition->setArgument(1, new Reference('doctrine.migrations.registry_loader'));
+
+            return;
+        }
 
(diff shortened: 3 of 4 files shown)


Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
```diff
@@ -277,6 +277,71 @@ public function testPrefersEntityManagerOverConnection() : void
...
+        return [
+            [null],
+            ['custom'],
+        ];
+    }
+
+    /**
+     * @dataProvider getEmNames
+     */
+    public function testPrefersMangerRegistry(?string $emName) : void
```
jmsche:
```suggestion
    public function testPrefersManagerRegistry(?string $emName) : void
```

Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
```diff
@@ -277,6 +277,71 @@ public function testPrefersEntityManagerOverConnection() : void
...
+
+        $container->compile();
+
+        $di = $container->get('doctrine.migrations.dependency_factory');
+
+        self::assertInstanceOf(DependencyFactory::class, $di);
+        self::assertSame($em, $di->getEntityManager());
+    }
+
+    public function testMangerRegistryThrowsExceptionOnUnknownEm() : void
```
jmsche:
```suggestion
    public function testManagerRegistryThrowsExceptionOnUnknownEm() : void
```

Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
```diff
@@ -277,6 +277,71 @@ public function testPrefersEntityManagerOverConnection() : void
         self::assertSame($em, $di->getEntityManager());
     }
 
+    /**
+     * @return array<string[]|null[]>
```
greg0ire:
```suggestion
     * @return array<array{string|null}>
```

Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
```diff
@@ -277,6 +277,71 @@ public function testPrefersEntityManagerOverConnection() : void
...
+     */
+    public function testPrefersManagerRegistry(?string $emName) : void
+    {
+        $config    = ['em' => $emName];
+        $container = $this->getContainer($config);
+
+        $em = $this->createMock(EntityManager::class);
+
+        $registryMock = $this->getMockBuilder(ManagerRegistry::class)
+            ->getMock();
```
greg0ire: Why not use `createMock` here too?

Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
```diff
@@ -277,6 +277,71 @@ public function testPrefersEntityManagerOverConnection() : void
...
+
+        $di = $container->get('doctrine.migrations.dependency_factory');
+
+        self::assertInstanceOf(DependencyFactory::class, $di);
+        self::assertSame($em, $di->getEntityManager());
+    }
+
+    public function testManagerRegistryThrowsExceptionOnUnknownEm() : void
+    {
+        $this->expectException(InvalidArgumentException::class);
```
greg0ire: Should be moved right before the line that is supposed to throw.

Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
```diff
@@ -277,6 +277,71 @@ public function testPrefersEntityManagerOverConnection() : void
...
+        $config    = ['em' => 'foo'];
+        $container = $this->getContainer($config);
+
+        $em = $this->createMock(EntityManager::class);
+
+        $registryMock = $this->getMockBuilder(ManagerRegistry::class)
+            ->getMock();
```
greg0ire: `createMock`?

composer.json
```diff
@@ -23,7 +23,7 @@
         "php": "^7.2",
         "symfony/framework-bundle": "~3.4|~4.0|~5.0",
         "doctrine/doctrine-bundle": "~1.0|~2.0",
-        "doctrine/migrations": "~3.0"
+        "doctrine/migrations": "^3.1@dev"
```
greg0ire: This makes the PR a draft, right? We are not going to merge as is, are we?

goetas: I not not see an issue with it. the `@dev` portion will be ignored by composer and for who requires this bundle, it will be seen as `^3.1`

Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
goetas: i've tried this, but PHPCS complains... this was the only syntax i've found that was making happy both phpstan and phpcs

goetas: but let me re try...

goetas: tried again, phpcs does not like it..

greg0ire: That might be because we use that super ancient version of our CS: https://github.com/doctrine/DoctrineMigrationsBundle/blob/aa15026788c2f5bfc4df899dfcd4d4612022fd1e/composer.json#L31

composer.json
greg0ire: So… they won't be able to install it then? Since `^3.1` can't be satisfied right now…

Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
greg0ire: See #365

composer.json
goetas: yes... ok, then we will wait for 3.1

Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
goetas: done

devnix:
Right now we are missing the possibility to have multiple migration folders separated by each entity manager, here is one suggestion of how to configure it: https://github.com/doctrine/DoctrineMigrationsBundle/issues/38#issuecomment-646665365

For example, in my case would be:
```yaml
doctrine_migrations:
    # default_connection: foo #(to specify a connection if there is not a "default" named connection there?)
    connections:
        default:
            migrations_paths:
                'DoctrineMigrations\Main': '%kernel.project_dir%/migrations/Main'
        geonames:
            migrations_paths:
                'DoctrineMigrations\Geonames': '%kernel.project_dir%/migrations/Geonames'
```

goetas:
@devnix what you have described is a feature request that goes out of scope for this change (and is much harder to implement)

loiclavoie:
@goetas Do you have a recommendation for those who depends on multiple connection/entity managers? We could put your answer in the 2 other issues concerning the missing --em/--connection feature in the latest upgrade.

goetas:
@loiclavoie That is an interesting question. 

Here there are two parts of the problem that needs to be solved:
1) generating migrations
2) running migrations


For (2 running migrations), the best thing to do IMO is to have migrations generated "per entity manager" and make sure to provide the right connection when running those migrations (that might be even trickier as migrations to not run queries, but just create a list of sql statements that later are executed by the executor, but it is not impossible).

If we accept the previous proposal , then to solve (1 generating migrations) we need to find a way to decide how to generate "entity manager migrations". 
What I could think about is: when running the `migration:diff command`, and the --namespace is passed, choose the entity manager associated with that namespace. (using the configuration style as what proposed by @devnix )

To me it seems feasible, but it might require some work.

devnix:
I've actually made wrappers for the diff, migrate and version commands. They sub call the original command, but with a `--em` parameter to specify which `DependencyFactory` should it use, and loading the configuration from a file with the same name of the entity manager. If I don't pass a `--em` parameter, I will iterate all the managers and run the command with the correct `DependencyFactory` each time.

For now, it seems to do the trick but it's a pain in the ass, and I must write a wrapper for each `doctrine:migrations` command I want to run.

If you think that the idea can help I can try to replicate this in a new project just for you. The configuration files are a little far from the proposal, I'm very limited by my lack of Doctrine's internals knowledge and the hacky approach.

composer.json
stof: well, you can still merge the PR. But you need to release the library before release the bundle.

goetas: @stof  exactly, that was my initial intention (cc @greg0ire )

devnix:
I wanted to make a reusable bundle of my wrapped commands to support multiple connections/migration_paths by just enabling the bundle and modifying the `doctrine_migrations.yaml` configuration, but this is the nearest thing I can get right now.

As I'm learning on the go, it seems like I can replace the `doctrine:migrations:*` commands with my wrappers and then call the original ones (can't extend them, as they are final classes), but I can't override the way `doctrine_migrations` TreeBuilder, so I have to append mine.

Now I'm almost there with a separated configuration, but I would have to test how all the possible parameters behave on each tree (and I didn't even tested XML files).

```yaml
doctrine_migrations:
    # Global values excepting "migrations_paths", "connection", "em"

doctrine_migrations_multiple_database:
    default_connection: default
    connections:
        default:
            migrations_paths:
                DoctrineMigrations\Main: '%kernel.project_dir%/migrations/Main'
        geonames:
            migrations_paths:
                DoctrineMigrations\Geonames: '%kernel.project_dir%/migrations/Geonames'
```

If I make it will be a very hacky (and maybe hard to catch up new versions of DoctrineMigrationsBundle) workaround, but easier to install and configure, as lots of people are demanding this feature several years ago.

I wish I would know more about the internals of Symfony and Doctrine. I'm learning a lot in the last couple of years by installing and hacking with the components in a legacy application and finally, migrate it fully into Symfony. But I feel I don't have the knowledge or the time in the short term to make a proper pull request.

If you don't mind it, I will post the repo link if you have any suggestion, as it will be far from perfect.

devnix:
I'm still testing a few things. As I mentioned, it is a very nasty way to overcome the problem, and I know that a good part of the code is not optimal at all, specially `DoctrineMigrationsMultipleDatabaseExtension::loadEntityManagerConfiguration()`, but any feedback is very welcome: https://github.com/AvaiBookSports/DoctrineMigrationsMultipleDatabaseBundle

I don't know if it can help someone to inspire a PR to DoctrineMigrationsBundle repository, I think that should be the end of the road, and not hacking it since 2012.

composer.json
greg0ire: >  you can still merge the PR

That sounds dangerous :stuck_out_tongue: Any lib should be releasable at any time IMO

goetas: this has been solved, doctrine/migrations 3.1 has been released

goetas:
This is ready for review again!

greg0ire (approved):

PR #374, opened 2020-11-04
Prevent deprecation with Symfony 5.1 and allow private services

smoench:
This PR tries to fix the following deprecation with Symfony 5.1:
```
1x: Since symfony/dependency-injection 5.1: The "Symfony\Component\DependencyInjection\ContainerInterface" autowiring alias is deprecated. Define it explicitly in your app if you want to keep using it. It is being referenced by the "App\Migrations\Factory\MigrationFactoryDecorator" service.
```
There is also the benefit of using private declared services.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -251,8 +251,9 @@ Here is an example on how to inject the service container into your migrations:
         services:
             Doctrine\Migrations\Version\DbalMigrationFactory: ~
             App\Migrations\Factory\MigrationFactoryDecorator:
-                decorates: Doctrine\Migrations\Version\DbalMigrationFactory
-                arguments: ['@App\Migrations\Factory\MigrationFactoryDecorator.inner', '@service_container']
+                arguments: ['@Doctrine\Migrations\Version\DbalMigrationFactory', '@service_container']
+                tags:
+                    - 'container.service_subscriber'
 
 
 .. code-block:: php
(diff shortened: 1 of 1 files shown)


Resources/doc/index.rst
```diff
@@ -251,8 +251,9 @@ Here is an example on how to inject the service container into your migrations:
         services:
             Doctrine\Migrations\Version\DbalMigrationFactory: ~
             App\Migrations\Factory\MigrationFactoryDecorator:
-                decorates: Doctrine\Migrations\Version\DbalMigrationFactory
```
stof, line 254: This is broken, as you don't decorate the service anymore, and so things won't work.

Resources/doc/index.rst
```diff
@@ -251,8 +251,9 @@ Here is an example on how to inject the service container into your migrations:
...
-                arguments: ['@App\Migrations\Factory\MigrationFactoryDecorator.inner', '@service_container']
```
stof, line 255: this code was **not** subject to the deprecation you mention, as it explicitly injects `@service_container` and does not rely on autowiring `Symfony\Component\DependencyInjection\ContainerInterface`

smoench: Okay, I had not set `@service_container` explicitly but this should be probably fixed with https://github.com/symfony/symfony/pull/38999

Resources/doc/index.rst
smoench: No, it is not broken. You need register `Doctrine\Migrations\Version\DbalMigrationFactory` as a service in your service configuration explicitly. There is no autoconfigure on the interface and you need define the migration factory implementation explicitly in your doctrine migrations config (see at line 248).

greg0ire:
Don't mind me, just debugging something with the CI at https://community.codecov.io/t/codecov-commit-status-check-missing-for-pr-when-using-codecov-action

PR #378, opened 2020-11-11
Maintenance work

greg0ire:


Command/DoctrineCommand.php
```diff
@@ -25,11 +25,11 @@
  */
 abstract class DoctrineCommand extends BaseCommand
 {
-    public static function configureMigrations(ContainerInterface $container, Configuration $configuration) : void
+    public static function configureMigrations(ContainerInterface $container, Configuration $configuration): void
     {
         $dir = $configuration->getMigrationsDirectory();
 
-        if (empty($dir)) {
+        if ($dir === null) {
```
SenseException, line 32: `$dir` could be an empty string in theory, right?

Command/DoctrineCommand.php
```diff
@@ -48,7 +48,7 @@ public static function configureMigrations(ContainerInterface $container, Config
             $pathPlaceholderArray = ['kernel.root_dir', 'kernel.cache_dir', 'kernel.logs_dir'];
 
             foreach ($pathPlaceholderArray as $pathPlaceholder) {
-                if (! $container->hasParameter($pathPlaceholder) || ! preg_match('/\%' . $pathPlaceholder . '\%/', $dir)) {
+                if (! $container->hasParameter($pathPlaceholder) || preg_match('/\%' . $pathPlaceholder . '\%/', $dir) === 0) {
```
SenseException, line 51: Wouldn't `!== 1` cover the error case too? It is currently not the same logic.

SenseException (changes requested):

Command/DoctrineCommand.php
greg0ire: Correct, but did we really mean to catch this case here?

Command/DoctrineCommand.php
greg0ire: Yes… but maybe I should add proper error handling (throw if it is `false`?). Given the value of `$pathPlaceholderArray`, I'd say this is never supposed to crash. Also, should `$pathPlaceholder` go through `preg_quote`? There is a `.` in it.

Command/DoctrineCommand.php
SenseException: No, definitely not here, but I can't evaluate the impact of this BC break. It looks like it won't have much impact.

Command/DoctrineCommand.php
SenseException: > maybe I should add proper error handling (throw if it is false?)

That would be awesome, but isn't this also a change of the current behaviour? Or do you want to `continue` in the `catch` while still in the 2.x branches?

> Also, should $pathPlaceholder go through preg_quote?

Good point. The values in `pathPlaceholder` are hardcoded, but because code never stays the same forever this could prevent problems.

greg0ire: > That would be awesome, but isn't this also a change of the current behaviour? Or do you want to continue in the catch while still in the 2.x branches?

I'm willing to bet it will never throw if nobody touches `$placeholderArray`, so I can't see how that would be a BC-break. I'm proposing this because we don't seem to use an assert library in this package.

:thinking: what would you prefer between that and a call to `assert()`?

Command/DoctrineCommand.php
greg0ire: Yeah I can't say I can evaluate it. I guess you're right, I should use a more complex condition just to be sure, the thing is, I think the original author would have picked `$dir === null` if `empty` had been forbidden earlier. It's what I believe but it's just that: a belief :sweat_smile:

Command/DoctrineCommand.php
SenseException: > thinking what would you prefer between that and a call to assert()?

Nah, I'm fine with `preg_quote`. You convinced me with the `preg_match` check.

Command/DoctrineCommand.php
SenseException: Let's go with our gut.

SenseException (approved):

Issue #360, opened 2020-08-13
Can not have any more no doctrine connection defined after upgrade to 3.0 ?


gggeek:
### BC Break Report
|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 3.0.1

#### Summary

I have a SF 4.4 app, originally created using the 'website-skeleton' composer package. As such, it does pull in orm-pack, which includes both doctrine and doctrine migrations.
The app's goal is to manage multiple databases, and as such it does not have a doctrine default connection set up.
In fact, it has no doctrine dbal connections defined at all (it sets up db connections on the fly when needed).
It also does not use doctrine migrations - even though it might in the future if the need arises...

#### Previous behaviour

The app worked fine.

#### Current behaviour

When running composer upgrade, doctrine migrations got bumped to 3.0.1. In the process, the Sf console command broke.
To fix that, I had to manually set the `migrations_paths` config in doctrine_migrations.yml.
However, as soon as I ran `bin/console`, I got an error: 

The service "doctrine.migrations.dependency_factory" has a dependency on a non-existent service "doctrine.dbal.default_connection".

#### Questions

1. Is it still possible to have Migrations installed without defining a db connection? If so, how can it be achieved? Reading around other bug reports, it seems that defining a custom default entity manager might do the trick, but it does not seem trivial for this scenario...

2. Given all the 'magic' happening with Symfony Packs, Flex and whatnot, I was slightly surprised to see a simple composer update break the app. Did I miss running any command that would have updated the config files automatically for me? If not, how comes that the standard set of dependencies for SF 4.4 apps does allow the bc-break-inducing Migrations 3.0 package to creep in?

gggeek:
PS: not sure if this is the same, but it feels related to issues #341 (and #38 )

erkens:
I had a similar problem with SF5.2 and I have made a small fix for this issue. Setting the connection configuration would do nothing when you are using the "default" entity manager.

PR #386, opened 2020-12-09
Fix usage of the "connection" configuration

erkens:
Setting an other database configuration then the "default" was not possible when you use the default entity manager. This change will check if the connection configuration is set before checking the entity manager. Because "connection" and "em" are mutually exclusive this wouldn't be a problem.

This could be a fix for #360, but I use SF5.2 for this.

### DIFF
--- a/DependencyInjection/CompilerPass/ConfigureDependencyFactoryPass.php
+++ b/DependencyInjection/CompilerPass/ConfigureDependencyFactoryPass.php
@@ -15,20 +15,20 @@ class ConfigureDependencyFactoryPass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container): void
     {
-        $preferredEm  = $container->getParameter('doctrine.migrations.preferred_em');
-        $diDefinition = $container->getDefinition('doctrine.migrations.dependency_factory');
+        $preferredEm         = $container->getParameter('doctrine.migrations.preferred_em');
+        $preferredConnection = $container->getParameter('doctrine.migrations.preferred_connection');
+        $diDefinition        = $container->getDefinition('doctrine.migrations.dependency_factory');
 
         $emID = sprintf('doctrine.orm.%s_entity_manager', $preferredEm ?? 'default');
 
-        if ($container->has($emID)) {
+        if ($preferredConnection === null && $container->has($emID)) {
             $container->getDefinition('doctrine.migrations.em_loader')
                 ->setArgument(0, new Reference($emID));
 
             $diDefinition->setFactory([DependencyFactory::class, 'fromEntityManager']);
             $diDefinition->setArgument(1, new Reference('doctrine.migrations.em_loader'));
         } else {
-            $preferredConnection = $container->getParameter('doctrine.migrations.preferred_connection');
-            $connectionId        = sprintf('doctrine.dbal.%s_connection', $preferredConnection ?? 'default');
+            $connectionId = sprintf('doctrine.dbal.%s_connection', $preferredConnection ?? 'default');
             $container->getDefinition('doctrine.migrations.connection_loader')
                 ->setArgument(0, new Reference($connectionId));
 
--- a/Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
+++ b/Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
@@ -260,6 +260,29 @@ class DoctrineMigrationsExtensionTest extends TestCase
         self::assertSame($conn, $di->getConnection());
     }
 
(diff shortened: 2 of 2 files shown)


goetas:
Hi, Is it possible to test this change?

erkens:
I have created a new test for this situation.

esbobkov:
Is this change ok?

greg0ire:
Please kindly squash your commits together. If you don't, we'll try to remember to do it for you but it's best if you save us this trouble.

How to do that?

1. `git rebase -i origin/3.0.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.
2. A window will show up with many lines, replace `pick` with `fixup` on every line but the first one
3. Close your editor, git should do its magic, and you should end up with one commit
4. Use `git push --force` to overwrite what you already push. Don't forget the `--force` option otherwise git will try to merge both things together.

greg0ire:
> Because "connection" and "em" are mutually exclusive this wouldn't be a problem.

Is that enforced anywhere? What happens when you configure both?

cas-technology:
> What happens when you configure both?

The preferred_connection is ignored and the default/set em is used. In this PR, the preferred_connection is used instead

greg0ire:
:thinking: wouldn't it be better to throw in that case?

erkens:
> 🤔 wouldn't it be better to throw in that case?

Why would you throw it? I want to use an other connection (preferred_connection) which isn't the same as my (default) entity_manager. So in the original code it checks if there is a prefered_em or a em with the name "default", and if either of these exists that will be used and NOT the preferred_connection.

greg0ire:
> Why would you throw it?

I would throw because I think it's better to tell the user that they are doing something that makes no sense than taking a guess at what they mean to do.

> I want to use an other connection (preferred_connection) which isn't the same as my (default) entity_manager.

How could they ever be the same? a connection and an entity manager are very different things, you cannot compare them, it's apple and oranges.

erkens:
> I would throw because I think it's better to tell the user that they are doing something that makes no sense than taking a guess at what they mean to do.

Ah, you mean when both preferred connection AND entitymanager is set. In that case I also like the idea of throwing that exception, but I think it is better to do that in an other PR, because that could break some instances I think.

> How could they ever be the same? a connection and an entity manager are very different things, you cannot compare them, it's apple and oranges.

I mean that the connection used in the default entitymanager isn't the same as the connection I want to use for the migration.

erkens:
Is there someone looking into this? It is quite annoying for me and my coworkers that I have to use a fork instead of the official package. I don't want to be rude or impatient, but if there is something wrong with my fix or if you don't want it, please tell me.

greg0ire:
It's fine to ping us, we get a lot of notifications, and sometimes personally I just clear them all. Too much noise :grimacing: 

> I think it is better to do that in an other PR, because that could break some instances I think.

If it breaks some instances, then it means your assumption that they are mutually exclusive is wrong, and that's what you are building your patch upon. IMO it would be best to throw, and if there happen to be some cases where it is in fact legitimate to use both options together, we can revert this PR, and rework it so that it takes this fact into account. If it's not the case, we will have gained better DX.

That's my opinion anyway, what do you think @goetas ?

erkens:
I mean that when someone gives both options for some "stupid" reason (because the only one that works with the current version is preferred_em, preferred_connection is ignored) it  will break if we throw an exception. Ofcourse it is good to throw that exception, but I didn't include it in this PR because this is just a bugfix and I don't like it when a bugfix will break existing code/scripts.

Throwing the exception when both options are used is an easy change, but this must also be updated in the documentation ofcourse.

greg0ire:
Okay. I retargeted to 3.1.x since now there are conflicts. Please fix those conflicts and force push.

erkens:
I think I should crreate a new PR for that, because I can't update my 3.0.x branch to 3.1.x because my fix is already used in several projects so these have to be upgraded to 3.1 first.

greg0ire:
You're right. Consider not naming it like the target branch this time, this is why it's best to use topic branches ;)

erkens:
> You're right. Consider not naming it like the target branch this time, this is why it's best to use topic branches ;)

yup, small mistake from my side, when thinking of it now, I don't know why I did it this way :)

erkens:
I'm not using this anymore, and I'm sorry that I didn't find time last year to fix it. Now I'm too long away from this code that I have to spend a lot of time to understand, fix and test it that it isn't worth it for me. So for now I will close this pull request (I will keep the fork and branch, so it isn't lost...)

PR #402, opened 2021-02-11
Assert container parameter types

greg0ire:
Not sure why PHPStan did not spot this before, but now it does.

### DIFF
--- a/Command/DoctrineCommand.php
+++ b/Command/DoctrineCommand.php
@@ -13,8 +13,12 @@ use Symfony\Component\DependencyInjection\ContainerAwareInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
 
+use function assert;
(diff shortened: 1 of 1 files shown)


Command/DoctrineCommand.php
```diff
@@ -137,4 +156,12 @@ private static function injectContainerToMigrations(ContainerInterface $containe
             $migration->setContainer($container);
         }
     }
+
+    public static function getStringParameter(ContainerInterface $container, string $name): string
```
andrew-demb:
```suggestion
    private static function getStringParameter(ContainerInterface $container, string $name): string
```

greg0ire: :thinking: why would I do that?

andrew-demb: Because this is not used anywhere except this class.

greg0ire: Oh sorry I read that backwards, fully agree!

SenseException (approved):

PR #399, opened 2021-02-08
documentation, incorrect migrations_paths example missing %kernel.project_dir%

LPodolski:
not correct config given in documentation:
```
doctrine_migrations:
  migrations_paths:
     'App\Migrations': 'src/App'
```

replaced to:
`'App\Migrations': '%kernel.project_dir%/src/Migrations'`

current is not correct and will give error: `Migrations directory "src/App" does not exist. ` tested in symfony 4.4 LTS

also added info about using `php bin/console make:migration` instead of `php bin/console doctrine:migrations:generate`

PR #403, opened 2021-02-18
Update index.rst adding project_dir to the first migration_path example

dant89:
An example using `%kernel.project_dir%` is useful because it doesn't break if you run console from a location other than the project root.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -45,10 +45,9 @@ application:
     doctrine_migrations:
         # List of namespace/path pairs to search for migrations, at least one required
         migrations_paths:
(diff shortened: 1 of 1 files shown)


stof:
I would actually replace the first example rather than adding another one. I don't think showing a case using relative path makes sense, as that's indeed relative to CWD which depends on how the code is run, making it unreliable.

SenseException (approved):

greg0ire:
Duplicate of #399

greg0ire:
Reopening since #399 has an unresolved comment

greg0ire:
Thanks @dant89 !

PR #407, opened 2021-03-07
Profiler collector

chr-hertel:
Taking over #143

toolbar icon only is visible when _new_ (red) or _unavailable_ (yellow) migrations are present:
Bildschirmfoto von 2021-03-12 16-56-02

more info on hover:
Bildschirmfoto von 2021-03-12 16-56-27

profiler panel:
Bildschirmfoto von 2021-03-12 16-56-44


goetas:
well, this is beautiful :heart:

Collector/MigrationsCollector.php
```diff
@@ -0,0 +1,87 @@
...
+        $statusCalculator = $this->dependencyFactory->getMigrationStatusCalculator();
+
+        $this->data['executed_migrations'] = $this->flattenExecutedMigrations($metadataStorage->getExecutedMigrations());
+        $this->data['available_migrations'] = $this->flattenAvailableMigrations($planCalculator->getMigrations());
+        $this->data['new_migrations'] = $this->flattenAvailableMigrations($statusCalculator->getNewMigrations());
+        $this->data['executed_unavailable_migrations'] = $this->flattenExecutedMigrations(
+            $statusCalculator->getExecutedUnavailableMigrations()
+        );
+
+        $configuration = $this->dependencyFactory->getConfiguration()->getMetadataStorageConfiguration();
```
goetas: if you want to print the storage class, it is more reliable  to use `getMetadataStorage()` (and is never null).

`getMetadataStorageConfiguration` is an optional class to implement since a storage might not be configurable and manage everything internally.

I would still use `getMetadataStorageConfiguration` to print  the table and column name.

Collector/MigrationsCollector.php
```diff
@@ -0,0 +1,100 @@
...
+        $connection = $this->dependencyFactory->getConnection();
+        $this->data['driver'] = get_class($connection->getDriver());
+        $this->data['name'] = $connection->getDatabase();
+
+        $this->data['namespaces'] = $configuration->getMigrationDirectories();
+    }
+
+    public function getName()
+    {
+        return 'migrations';
```
stof: should be `doctrine_migrations` instead, to have proper namespacing.

Resources/views/Collector/icon.svg
```diff
@@ -0,0 +1,7 @@
+<svg viewBox="9.954 10.02 33.268 32.669" width="33.268" height="32.669">
```
stof: Why such weird viewbox instead of using a `0 0 24 24` like all the existing profiler icons ?

Resources/views/Collector/icon.svg
```diff
@@ -0,0 +1,7 @@
...
+    <polygon fill="#AAA" points="10.204000473022461 33.944000244140625 20.1299991607666 33.891998291015625 21.15399932861328 37.93600082397461 22.152000427246094 33.917999267578125 27.457000732421875 33.970001220703125 26.195999145507812 39.43299865722656 32.6150016784668 39.43299865722656 37.97100067138672 39.43299865722656 36.59400177001953 33.99700164794922 43.00199890136719 33.891998291015625 43.00199890136719 42.63600158691406 10.125 42.68899917602539" />
+    <polygon fill="#AAA" points="9.954000473022461 10.020000457763672 42.803001403808594 10.020000457763672 42.803001403808594 18.70800018310547 32.88399887084961 18.684999465942383 31.909000396728516 14.734999656677246 31.000999450683594 18.66200065612793 25.56599998474121 18.70800018310547 26.889999389648438 13.24899959564209 21.34000015258789 13.24899959564209 15.060999870300293 13.24899959564209 16.43600082397461 18.70800018310547 10.00100040435791 18.684999465942383" />
```
stof: Creating the icon aligned on a 24px grid as the core profiler icons could help avoiding such weird coordinates.

Resources/views/Collector/migrations.html.twig
```diff
@@ -0,0 +1,192 @@
+{% extends app.request.isXmlHttpRequest ? '@WebProfiler/Profiler/ajax_layout.html.twig' : '@WebProfiler/Profiler/layout.html.twig' %}
```
stof: not necessary, as this panel does not perform any hack ajax-based loading like the DoctrineBundle one.

Collector/MigrationsCollector.php
```diff
@@ -0,0 +1,100 @@
...
+    {
+        return 'doctrine_migrations';
+    }
+
+    public function getData()
+    {
+        return $this->data;
+    }
+
+    public function reset()
```
stof, line 66: isn't that exactly the implementation in the parent class already ?

Resources/views/Collector/migrations.html.twig
```diff
@@ -0,0 +1,192 @@
...
+
+{% block toolbar %}
```
stof, line 5: As this panel mostly brings value when you have unavailable or new migrations, which won't be the case 99% of the cases, maybe it should hide itself from the toolbar for other cases to save space, as done for many core panels ?

Collector/MigrationsCollector.php
```diff
@@ -0,0 +1,100 @@
...
+        $this->dependencyFactory = $dependencyFactory;
+    }
+
+    public function collect(Request $request, Response $response, \Throwable $exception = null)
+    {
+        $metadataStorage = $this->dependencyFactory->getMetadataStorage();
+        $planCalculator = $this->dependencyFactory->getMigrationPlanCalculator();
+        $statusCalculator = $this->dependencyFactory->getMigrationStatusCalculator();
+
+        $availableMigrations = $planCalculator->getMigrations();
```
stof, line 33: What is the performance cost of collecting all this information about migrations ? To me, this will bring value only if the status is not green (in which case there is a good chance chance that you get an exception page because the ORM tries to query non-existent fields or tables).
If this takes time to collect, the cost/benefit ratio might not be good enough for DX.

Resources/views/Collector/migrations.html.twig
```diff
@@ -0,0 +1,192 @@
...
+            </tr>
+        {% endfor %}
+
+        {% set unavailable = [] %}
+        {% for migration in collector.data.executed_unavailable_migrations %}
+            {% set unavailable = unavailable|merge([migration.version]) %}
+        {% endfor %}
+
+        {% for migration in collector.data.executed_migrations|reverse %}
+            {% set is_unavailable = migration.version in unavailable %}
```
stof: isn't this just `is_unavailable = migration in collector.data.executed_unavailable_migrations`, avoiding to build the `unavailable` list ?

Resources/views/Collector/migrations.html.twig
```diff
@@ -0,0 +1,192 @@
...
+            <tr>
+                <td class="font-normal">
+                    {% if is_unavailable %}
+                        {{ migration.version }}
+                    {% else %}
+                        {% set version = migration.version|split('\\') %}
```
stof: the variable `version` looks weird to me, as it is not a version but an array.

Resources/views/Collector/migrations.html.twig
```diff
@@ -0,0 +1,192 @@
...
+                        {% set file = collector.data.namespaces[version|first] ~ '/' ~ version|last ~ '.php' %}
```
stof: what if there are multiple `\` ?

stof (commented):
I'm still questioning the value of having that in the profiler, as it is not really about giving info about the profiled request.

chr-hertel:
@stof the idea was to improve DX because I often saw developers having issues after switching branches and having a panel might help to see unexecuted or unavailable migrations. the second part was to access the status info in the panel with link to the migration files - which allows to get fast insights w/o switching to CLI.

this is how the toolbar might help while running in migration related errors:
Bildschirmfoto von 2021-03-08 19-37-05

i like your idea of only rendering the toolbar if there are new or unavailable migrations and we could get rid off performance implications by switching to rendering a controller in the panel instead of storing the data in the profile. WDYT?

Collector/MigrationsCollector.php
chr-hertel: are you referring to `Symfony\Bundle\FrameworkBundle\DataCollector\AbstractDataCollector`? using `Symfony\Component\HttpKernel\DataCollector\DataCollector` here instead due to the 3.x/4.x support.
or do i miss sth?

wouterj:
> i like your idea of only rendering the toolbar if there are new or unavailable migrations

:+1: That also matches the standards with the core profiler tabs. I would also invert the icon (only show the arrows, not the background) and only show the number of new or unavailable migrations (`2` instead of `2 / 2`), to match the other information.

> we could get rid off performance implications by switching to rendering a controller in the panel instead of storing the data in the profile

I don't think this is a good idea. I would expect the profiler to show me the state at the time of the request, not at the current time (maybe migrations are executed between the request and the moment I'm looking at the profiler page).

---

> the idea was to improve DX because I often saw developers having issues after switching branches and having a panel might help to see unexecuted or unavailable migrations

I also have this experience with less Symfony-knowledged developers in our team (e.g. front-end or app developers). After switching branches, they'll see confusing database errors, which requires teaching them "if the error contains SQL, just run migrations". Such a panel would help identifying these issues much more clearer (but of course, it shouldn't influence performance too much).

Collector/MigrationsCollector.php
chr-hertel: reduced the profiler data to the names only, loading data using a twig extension.
should we add an config key to disable the profiler, like the original PR proposed?

Collector/MigrationsCollector.php
dbrumann: As far as I can tell, the abstract DataCollector does not implement this. So, looks good to me

dbrumann (approved):
I just played around with this in the symfony demo app using both mysql and sqlite. I didn't encounter any issues and really like the DX this brings.

I can't say much about the performance concerns. Adding a toggle to the config sounds like a good fix, but could be done in a follow up as far as I am concerned.

goetas:
@chr-hertel  This looks good IMO. Regarding the performance impact, most of the time is spent in loading the list of migrations from DB and later instantiating the related classes and sorting them, thus what is done in the twig extension is not that useful to save some ms, since most of the work has been already done by the collector.

I'm referring to this lines in the collector:

```php
$this->data['available_migrations'] = $this->flattenMigrations($planCalculator->getMigrations());
$this->data['executed_migrations'] = $this->flattenMigrations($metadataStorage->getExecutedMigrations());
$this->data['new_migrations'] = $this->flattenMigrations($statusCalculator->getNewMigrations());
$this->data['unavailable_migrations'] = $this->flattenMigrations($statusCalculator->getExecutedUnavailableMigrations());
```

They take 90% (if not more)... but they have to stay there...

I've created https://github.com/chr-hertel/DoctrineMigrationsBundle/pull/1 as a pull request to your branch that moves the operations to the collector (and does not require special twig extensions).

If you agree with my changes in https://github.com/chr-hertel/DoctrineMigrationsBundle/pull/1 you can merge them, I think that this is ready to go.

Collector/MigrationsCollector.php
goetas: @stof are you referring to the cost of getting the data from doctrine migrations or the size on disk of the collected profile?

chr-hertel:
Thanks for your help @goetas! Just merged your PR and added a config flag to disable profiler.

goetas (approved):
To me this looks very good! The extra flag to disable the profiles makes it also better since ppl that have a lot of migrations can disable it!

derrabus (approved):

wouterj:
Just a thought: if performance for large applications (or large migration sets) is compromised by this feature, would it make sense to disable it by default? And then ship a recipe that enables it ("by default")? This way, we have a good visibility of the feature, but we don't force existing applications to figure out where the performance impact comes from.

goetas:
@wouterj 
> Just a thought: if performance for large applications (or large migration sets) is compromised by this feature, would it make sense to disable it by default? And then ship a recipe that enables it ("by default")? This way, we have a good visibility of the feature, but we don't force existing applications to figure out where the performance impact comes from.

I think that this is a great idea! It will make the bundle 100% backward compatible and there will be no cases in which the profiler is enabled without "user consent". @chr-hertel  what do you think?

goetas:
A question for @alcaeus  @greg0ire  is about testing... should this feature have tests? Is it a common practice to have unit tests for profiles? As example checking that the data collected by `MigrationsCollector` are correct?

greg0ire:
There is some complexity in the `flatten*` methods, so it would make sense to have unit tests for `MigrationsCollector`, although I don't have an example to share. @chr-hertel have you considered writing tests (not necessarily unit tests)?

chr-hertel:
great input! thanks!

next steps
- [x] profiler disabled by default
- [x] symfony recipe change for 3.1
- [x] tests

will tackle those the next days! :)

chr-hertel:
Tackled all proposed/required changes and drafted a corresponding PR https://github.com/symfony/recipes/pull/927

dbrumann (approved):

goetas (approved):
thanks a lot @chr-hertel !

greg0ire:
Thanks @chr-hertel !

PR #416, opened 2021-04-10
Debug toolbar breaks when executed migrations list is empty

HypeMC:
If there are no previously executed migrations the `last` function returns `false` which breaks the toolbar.

image


derrabus:
👀 @chr-hertel

chr-hertel (approved):
LGTM, thanks! :)

Issue #417, opened 2021-04-20
Profiler collector is way too expensive


Tetragramat:
I have relatively small project with 18 migrations, but profiler collector takes about 40% of the execution time.
Profiler collector is nice addition to DX, but it reduces DX with way too much slower response time.

With profiler collector:
![with](https://user-images.githubusercontent.com/7849117/115430000-4409e480-a204-11eb-9df6-0aab24fc7e9c.png)

Without:
![without](https://user-images.githubusercontent.com/7849117/115430096-5b48d200-a204-11eb-8cea-40690ac5f27f.png)

goetas:
@Tetragramat i think that the profiler is disabled by default. did you investigate which function calls are particularly expensive?

Tetragramat:
Too many calls of `Doctrine\DBAL\Schema\AbstractSchemaManager::listTableNames` and `Doctrine\DBAL\Schema\AbstractSchemaManager::listTableColumns` which contain slow queries.

```
^ "default: 26.00 MiB - 33 ms"
^ "Doctrine\DBAL\Schema\AbstractSchemaManager::listTableNames"
^ "default: 28.00 MiB - 28 ms"
^ "Doctrine\DBAL\Schema\AbstractSchemaManager::listTableNames"
^ "default: 28.00 MiB - 34 ms"
^ "Doctrine\DBAL\Schema\AbstractSchemaManager::listTableColumns"
^ "default: 28.00 MiB - 28 ms"
^ "Doctrine\DBAL\Schema\AbstractSchemaManager::listTableNames"
^ "default: 28.00 MiB - 28 ms"
^ "Doctrine\DBAL\Schema\AbstractSchemaManager::listTableNames"
^ "default: 28.00 MiB - 34 ms"
^ "Doctrine\DBAL\Schema\AbstractSchemaManager::listTableColumns"
^ "default: 28.00 MiB - 28 ms"
^ "Doctrine\DBAL\Schema\AbstractSchemaManager::listTableNames"
^ "default: 28.00 MiB - 27 ms"
^ "Doctrine\DBAL\Schema\AbstractSchemaManager::listTableNames"
^ "default: 28.00 MiB - 37 ms"
^ "Doctrine\DBAL\Schema\AbstractSchemaManager::listTableColumns"
```

goetas:
Hmm, that sounds like a bug, that function should be called only once.. 🤔

chr-hertel:
[...]

PR #421, opened 2021-05-17
Reduce profiler costs

chr-hertel:
Reducing profiler costs by skipping usage of StatusCalculator

addressing #417

### DIFF
--- a/Collector/MigrationsCollector.php
+++ b/Collector/MigrationsCollector.php
@@ -5,10 +5,6 @@ declare(strict_types=1);
 namespace Doctrine\Bundle\MigrationsBundle\Collector;
 
 use Doctrine\Migrations\DependencyFactory;
-use Doctrine\Migrations\Metadata\AvailableMigration;
-use Doctrine\Migrations\Metadata\AvailableMigrationsList;
-use Doctrine\Migrations\Metadata\ExecutedMigration;
-use Doctrine\Migrations\Metadata\ExecutedMigrationsList;
 use Doctrine\Migrations\Metadata\Storage\TableMetadataStorageConfiguration;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
@@ -36,10 +32,10 @@ class MigrationsCollector extends DataCollector
         $availableMigrations = $planCalculator->getMigrations();
 
         $this->data['available_migrations_count'] = count($availableMigrations);
-        $unavailableMigrations = $this->filterUnavailableMigrations($executedMigrations, $availableMigrations);
+        $unavailableMigrations = $executedMigrations->unavailableSubset($availableMigrations);
         $this->data['unavailable_migrations_count'] = count($unavailableMigrations);
 
-        $newMigrations = $this->filterNewMigrations($availableMigrations, $executedMigrations);
+        $newMigrations = $availableMigrations->newSubset($executedMigrations);
         $this->data['new_migrations'] = $this->flattener->flattenAvailableMigrations($newMigrations);
         $this->data['executed_migrations'] = $this->flattener->flattenExecutedMigrations($executedMigrations, $availableMigrations);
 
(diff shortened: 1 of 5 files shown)


Collector/MigrationsFlattener.php
```diff
@@ -11,30 +11,30 @@
 
 class MigrationsFlattener
 {
-    public function flattenAvailableMigrations(AvailableMigrationsList $migrationsList, ?ExecutedMigrationsList $executedMigrations = null): array
+    public function flattenNewMigrations(AvailableMigrationsList $migrationsList, ExecutedMigrationsList $executedMigrations): array
```
goetas: overall looks very good, the only thing that confused me is the name of this method.

`flattenNewMigrations` gives the impression to flatten a list on new migrations, but the method is first calculating the list of new migrations using the executed and available migrations, and then is flattening them.

would it make sense to give a better name to the method or to extract the process of calculating the new migrations in a private method as part of the data collector class?

chr-hertel: yup, that's a good idea. wasn't sure about this anyways because it duplicates logic of `Doctrine\Migrations\Version\CurrentMigrationStatusCalculator::getNewMigrations`, but that would bring a refactoring to the migrations lib which felt like a scope creep.
what do you think about that?

goetas: yea, i understand, but personally in think that is fine.

goetas (approved):

SenseException (approved):

Collector/MigrationsCollector.php
```diff
@@ -67,4 +72,26 @@ public function reset()
...
         $this->data = [];
     }
+
+    private function filterUnavailableMigrations(
+        ExecutedMigrationsList $executedMigrations,
+        AvailableMigrationsList $availableMigrations
+    ): ExecutedMigrationsList {
+        $unavailableMigrations = array_filter($executedMigrations->getItems(), static function (ExecutedMigration $migration) use ($availableMigrations) {
+            return !$availableMigrations->hasMigration($migration->getVersion());
+        });
```
greg0ire: Wouldn't these 3 lines be better in `ExecutedMigrationsList`?

```php
public function unavailableSubset(AvailableMigrationsList $availableMigrations): self
{
    return new self(array_filter($this->getItems(), static function (ExecutedMigration $migration) use ($availableMigrations)       {
        return !$availableMigrations->hasMigration($migration->getVersion());
    }));
}
```

Collector/MigrationsCollector.php
```diff
@@ -67,4 +72,26 @@ public function reset()
...
+    private function filterNewMigrations(
+        AvailableMigrationsList $availableMigrations,
+        ExecutedMigrationsList $executedMigrations
+    ): AvailableMigrationsList {
+        $newMigrations = array_filter($availableMigrations->getItems(), static function (AvailableMigration $migration) use ($executedMigrations) {
+            return !$executedMigrations->hasMigration($migration->getVersion());
+        });
+
+        return new AvailableMigrationsList($newMigrations);
+    }
```
greg0ire: Same here, and it will be easier to unit test those.

Collector/MigrationsCollector.php
chr-hertel: Good idea, thanks, done here: https://github.com/doctrine/migrations/pull/1174

greg0ire: Ah I didn't realize that it was in a different package. @goetas , would you be fine with releasing 3.2.0 soon? Otherwise, we can merge this as is, and migrate to the new methods once 3.2.0 is out.

goetas:
@chr-hertel https://github.com/doctrine/migrations/pull/1174 has been released as `3.2.0`. now you can update your pull request to rely on the feature implemented in the library.

goetas (changes requested):
@chr-hertel doctrine/migrations#1174 has been released as `3.2.0`. Bow you can update your pull request to rely on the feature implemented doctrine/migrations#1174.

goetas (approved):

greg0ire:
Thanks @chr-hertel !

PR #425, opened 2021-06-04
Fix build

greg0ire:


Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
```diff
@@ -419,6 +420,7 @@ private function getContainer(array $config, ?array $dbalConfig = null, ?array $
 
         $container->getDefinition('doctrine.migrations.dependency_factory')->setPublic(true);
         $container->getDefinition('doctrine.migrations.configuration')->setPublic(true);
+        $container->addCompilerPass(new CacheCompatibilityPass());
```
greg0ire: At first I went with a `class_exists` but then thought we might as well bump the dependency on `doctrine/doctrine-bundle`

composer.json
```diff
@@ -22,7 +22,7 @@
     "require": {
         "php": "^7.2|^8.0",
         "symfony/framework-bundle": "~3.4|~4.0|~5.0",
-        "doctrine/doctrine-bundle": "~1.0|~2.0",
+        "doctrine/doctrine-bundle": "^2.4.1",
```
SenseException: Is it okay to do this in a patch version? It's no BC break but will this constraint exclude many updates in projects?

greg0ire: I was not sure about this, hence my comment above. Since you're not sure either, let's revert to `class_exists`

derrabus (approved):

PR #428, opened 2021-06-19
Ignore missing connection in MigrationsCollector

### LINKED ISSUE #423: MigrationCollector always requires database connection
Maybe it's just me but I find it annoying that the out of the box Symfony website-skeleton requires an actual database connection for the migrations bundle even if no database is currently being used.  I also see plenty of help requests from developers who are stymied when they encounter this sort of error.  Sure they should be able to work around the issue but it is still annoying.

The problem is that the migrations bundle profile data collector always checks for migrations tables which in turn requires a valid connection:
```
# Doctrine\Bundle\MigrationsBundle\Collector\MigrationsCollector
public function collect(Request $request, Response $response, \Throwable $exception = null)
    {
        // Add a data template
        $this->data = [
            'available_migrations' => null,
            'executed_migrations' => null,
            'new_migrations' => null,
            'unavailable_migrations' => null,
            'storage' => null,
            'table' => null,
            'column' => null,
            'driver' => null,
            'name' => null,
            'namespaces' => null,
        ];

        $metadataStorage = $this->dependencyFactory->getMetadataStorage();
        $planCalculator = $this->dependencyFactory->getMigrationPlanCalculator();
        $statusCalculator = $this->dependencyFactory->getMigrationStatusCalculator();

        // Add a try catch
        try {
            $executedMigrations  = $metadataStorage->getExecutedMigrations();
        }
        catch (\Exception $e) { 
            // Basically ignore the data collector
            return;
        }
```
The above proposed fix catches the connection exception and returns an empty data collector.  I suppose there are more sophisticated ways of doing this and using exceptions for flow control is frowned upon but in this case I think it is fine.

Is this the sort of thing for which a PR might be welcomed?  Just a minor improvement for the Developer's UX.

dbrumann:
Closes #423 

* Ignores connection error during collection
* Does not render toolbar or menu entry, when collector was not triggered

### DIFF
--- a/Collector/MigrationsCollector.php
+++ b/Collector/MigrationsCollector.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\Bundle\MigrationsBundle\Collector;
 
+use Doctrine\DBAL\Exception;
 use Doctrine\Migrations\DependencyFactory;
 use Doctrine\Migrations\Metadata\Storage\TableMetadataStorageConfiguration;
 use Symfony\Component\HttpFoundation\Request;
@@ -32,7 +33,17 @@ class MigrationsCollector extends DataCollector
         $metadataStorage = $this->dependencyFactory->getMetadataStorage();
         $planCalculator = $this->dependencyFactory->getMigrationPlanCalculator();
 
-        $executedMigrations  = $metadataStorage->getExecutedMigrations();
+        try {
+            $executedMigrations = $metadataStorage->getExecutedMigrations();
+        } catch (Exception $dbalException) {
+            $this->dependencyFactory->getLogger()->error(
+                'error while trying to collect executed migrations',
+                ['exception' => $dbalException]
+            );
+
+            return;
+        }
+
         $availableMigrations = $planCalculator->getMigrations();
 
         $this->data['available_migrations_count'] = count($availableMigrations);
--- a/Resources/views/Collector/migrations.html.twig
+++ b/Resources/views/Collector/migrations.html.twig
@@ -3,61 +3,65 @@
 {% import _self as helper %}
 
 {% block toolbar %}
-    {% set unavailable_migrations = collector.data.unavailable_migrations_count %}
-    {% set new_migrations = collector.data.new_migrations|length %}
-    {% if unavailable_migrations > 0 or new_migrations > 0 %}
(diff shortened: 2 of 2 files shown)


goetas:
Hi @dbrumann , thanks for your PR. Your change looks good, but I would like to wait for the changes in https://github.com/doctrine/DoctrineMigrationsBundle/pull/421 by  @chr-hertel  and then to merge this (they are working on the same files and chances of git conflicts are high :( )

dbrumann:
I understand. I would be fine with closing this PR and incorporating it there instead. No hurry from my side. I just stumbled upon this via a StackOverflow question.

Collector/MigrationsCollector.php
```diff
@@ -29,7 +30,13 @@ public function collect(Request $request, Response $response, \Throwable $except
         $planCalculator = $this->dependencyFactory->getMigrationPlanCalculator();
         $statusCalculator = $this->dependencyFactory->getMigrationStatusCalculator();
 
-        $executedMigrations  = $metadataStorage->getExecutedMigrations();
+        try {
+            $executedMigrations  = $metadataStorage->getExecutedMigrations();
+        } catch (Exception $dbalException) {
+            // Ignore connection error and do not collect any data
```
greg0ire: Should this at least be logged with
```
$this->dependencyFactory->getLogger()->error(
    'error while trying to collect executed migrations',
    ['exception' => $dbalException]
);
```

Alternatively, have you considered collecting the exception for later display inside the collector?

greg0ire (commented):
Please add some tests :pray:

Collector/MigrationsCollector.php
dbrumann: Yes, I was considering logging, but didn't want to add the additional dependency into the class.

goetas:
@dbrumann would you mind rebasing this now that @chr-hertel 's changes (https://github.com/doctrine/DoctrineMigrationsBundle/pull/421) have been merged?

dbrumann:
Yes, I will have a look at it later this week

chr-hertel (approved):

goetas (approved):

greg0ire:
Since this is an improvement (described as such in the closed issue), it should target 3.2.x

greg0ire:
Thanks @dbrumann !

PR #435, opened 2021-07-21
Remove duplicate collector calls

Jean-Beru:
When a sub-request is done (eg: with `controller` Twig function), the collector is called twice.

This PR will avoid additional database queries in this case.


greg0ire (approved):

greg0ire:
Maybe this should target 3.2.x since it's not a bugfix? :thinking:

SenseException (approved):
It's neither a feature.

greg0ire:
Oh well… let's merge. Thanks @Jean-Beru !

PR #453, opened 2021-11-10
Require PHPUnit >= 8

greg0ire:
We have dropped PHP 7.1, which means we can drop PHPUnit 7, which will
in turn result in the prefer-lowest job being able to generate a
coverage file, instead of warning that "No coverage driver is
available".

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "doctrine/migrations": "^3.2"
     },
     "require-dev": {
-        "phpunit/phpunit": "^7.0|^8.0|^9.0",
+        "phpunit/phpunit": "^8.0|^9.0",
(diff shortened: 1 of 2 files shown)


derrabus (approved):

composer.json
```diff
@@ -26,7 +26,7 @@
         "doctrine/migrations": "^3.2"
     },
     "require-dev": {
-        "phpunit/phpunit": "^7.0|^8.0|^9.0",
+        "phpunit/phpunit": "^8.0|^9.0",
```
derrabus, line 29: We could also use `^8.5.21|^9.5` here. We don't gain much from running the `lowest` job with PHPUnit 8.0.0.

PR #456, opened 2021-11-11
Reuse Symfony unstable workflow

greg0ire:


composer.json
```diff
@@ -21,11 +21,10 @@
     ],
     "require": {
         "php": "^7.2|^8.0",
-        "symfony/framework-bundle": "~3.4|~4.0|~5.0|~6.0",
```
derrabus: Should be reverted. Otherwise Composer would refuse to install the bundle with Symfony 6 again.

psalm.xml
```diff
@@ -15,13 +15,4 @@
...
     </projectFiles>
-
-    <issueHandlers>
-        <!-- This check is confused by Symfony's "mixed" property types. -->
-        <ReservedWord>
-            <errorLevel type="suppress">
-                <directory name="." />
-            </errorLevel>
-        </ReservedWord>
-    </issueHandlers>
```
derrabus: As soon as Symfony 6 is stable, Psalm will fail again. Let's keep the issue handler in here until we have a better solution.

derrabus (changes requested):

composer.json
greg0ire: It's just here to demonstrate how it's possible to test versions that are not supported officially. Will be reverted 👍

psalm.xml
greg0ire: I thought this was because we were using bleeding edge Psalm?

derrabus: I thought this was a bleeding edge Psalm issue at first, but it turns out that stable Psalm has the same issue. :-/

greg0ire:
@derrabus I just pushed with `6.0.x-dev` as a constraint, but this only results in the framework-bundle being installed in that version. I think the case of Symfony is kind of special. Is there a trick I could use in the workflow to have all symfony components on 6.0.x-dev, and all other packages in a stable version?

derrabus:
> Is there a trick I could use in the workflow to have all symfony components on 6.0.x-dev, and all other packages in a stable version?

If you add `flex` to the `tools` parameter of the Setup PHP step, this should do the trick:

```sh
composer config extra.symfony.require '6.0.*'
```

But I just noticed that `doctrine/migrations` is blocking `symfony/console` 6. We need to solve that first, I guess.

greg0ire:
Then I should make the upstream workflow Symfony-specific, right? This wouldn't work for laminas, would it? Another solution would be to require `symfony/symfony 6.0.x`, but I guess cloning the whole repo might be slow. Let me try.

greg0ire:
That does not work, because I would have to drop `symfony/framework-bundle`

greg0ire:
@derrabus I think I have managed to do something that behaves correctly (as in it allows us to detect that we should focus on `doctrine/migrations` first)

greg0ire:
ping @derrabus :slightly_smiling_face:

derrabus:
On the Symfony 6 workflow, Composer installs the `symfony/symfony` monorepo now which is certainly not what we want. I wonder why that happens.

.github/workflows/continuous-integration.yml
```diff
@@ -15,3 +15,10 @@ jobs:
     uses: "doctrine/.github/.github/workflows/continuous-integration.yml@1.2.1"
     with:
       php-versions: '["7.2", "7.3", "7.4", "8.0"]'
+
+  phpunit-with-symfony-6:
+    name: "PHPUnit with Symfony 6.0.x-dev"
+    uses: "doctrine/.github/.github/workflows/continuous-integration-symfony-unstable.yml@1.4.0"
+    with:
+      symfony-version-constraint: "6.0.x-dev"
+      extra-requirements: "symfony/framework-bundle 6.0.x-dev symfony/symfony 6.0.x-dev"
```
derrabus:
```suggestion
      symfony-version-constraint: "6.0.*"
```

Can we try this?

greg0ire:
> On the Symfony 6 workflow, Composer installs the symfony/symfony monorepo now which is certainly not what we want. I wonder why that happens. 

Well spotted! And also… WTF :sweat_smile: ? 

EDIT: Oh… it's just because I wrote `symfony/symfony` in the requirements :facepalm: … a remnant of my previous attempts. Anyway, as you pointed out, using extra requirements is not needed here since we are allowing Symfony 6 already.

PR #471, opened 2022-02-21
Document profiler overhead

94noni:
After upgrading the bundle recipe on a local app
I saw the new profiler icon and also some queries overhead (~10queries)

Ref https://github.com/symfony/recipes/blob/9a7f5bcebbda2f5244c4be4133a6bd181080cbea/doctrine/doctrine-migrations-bundle/3.1/config/packages/doctrine_migrations.yaml#L6

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -83,6 +83,9 @@ application:
         # Whether or not to wrap migrations in a single transaction.
         transactional: true
 
+        # Whether or not to enable the profiler collector. This add some queries overhead.
+        # enable_profiler: false
+
         services:
             # Custom migration sorting service id
             'Doctrine\Migrations\Version\Comparator': ~
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -83,7 +83,7 @@ application:
         # Whether or not to wrap migrations in a single transaction.
         transactional: true
 
-        # Whether or not to enable the profiler collector. This add some queries overhead.
+        # Use profiler to calculate and visualize migration status.
         # enable_profiler: false
 
         services:
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -83,7 +83,7 @@ application:
         # Whether or not to wrap migrations in a single transaction.
         transactional: true
 
-        # Use profiler to calculate and visualize migration status.
+        # Whether or not to enable the profiler collector to calculate and visualize migration status. This adds some queries overhead.
         # enable_profiler: false
 
         services:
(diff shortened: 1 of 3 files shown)


Resources/doc/index.rst
```diff
@@ -83,6 +83,9 @@ application:
         # Whether or not to wrap migrations in a single transaction.
         transactional: true
 
+        # Whether or not to enable the profiler collector. This add some queries overhead.
```
greg0ire:
```suggestion
        # Whether or not to enable the profiler collector. This adds some queries overhead.
```

greg0ire (commented):
I think this should be put in sync with https://github.com/doctrine/DoctrineMigrationsBundle/blob/453b0979579c4ab214a37b28d0b3cc8b471ac072/DependencyInjection/Configuration.php#L150-L153

Resources/doc/index.rst
94noni: @greg0ire indeed I can use the same sentence `Use profiler to calculate and visualize migration status.`

do I need to target v3.1? and keep there value `false`

thx

greg0ire: The correct branch to target is 3.2, as the last release is 3.2.2

greg0ire:
> Update index.rst

You are going to edit several files, so might as well use something higher level, like "Document profiler overhead"

SenseException:
> You are going to edit several files

I don't understand what the scope is for this PR.

greg0ire:
IMO it should be:
- Resources/doc/index.rst
- DoctrineMigrationsBundle/DependencyInjection/Configuration.php

That way we have some sync between both files.

94noni:
the main goal is to document this configuration that is not present in the https://symfony.com/bundles/DoctrineMigrationsBundle/current/index.html

the current readme change can be reverted and if accepted, done in another PR

greg0ire:
> the current readme change can be reverted and if accepted, done in another PR

It's fine to me, but to make things perfectly clear, I would expect your addition about the overhead to appear in `DoctrineMigrationsBundle/DependencyInjection/Configuration.php` as well. Right now I no longer even see it in the diff. All I see is that you synced `Resources/doc/index.rst` which is great, but now I'm waiting for you to add your sentence about the overhead in both files.

94noni:
@greg0ire it is absolutely not what I have understood reading (and rereading) your comments
sorry about that, I will rewrite soon with what I had in mind first

94noni:
@greg0ire hi, here are the changes I propose, thx!

greg0ire (approved):
To be squash merged

greg0ire:
Thanks @94noni !

PR #474, opened 2022-03-07
Added condition to check that we have an active connection

### LINKED ISSUE #473: MigrationsCollector still requires a active connection
Collector still requires an configure connection.
When starting new project and database is not yet configured, MigrationsCollector is trying to connect to database and because of PDO timeout, page feels are hanged.

Ref.:
https://github.com/doctrine/DoctrineMigrationsBundle/pull/428
https://github.com/doctrine/DoctrineMigrationsBundle/issues/423

oleg-andreyev:
Fixes #473

### DIFF
--- a/Collector/MigrationsCollector.php
+++ b/Collector/MigrationsCollector.php
@@ -35,6 +35,9 @@ class MigrationsCollector extends DataCollector
         }
 
         $connection = $this->dependencyFactory->getConnection();
+        // before collecting migrations, we need to make sure that Application has
+        // at least opened a connection, otherwise it's not guaranteed that connection
+        // is even configured
         if (! $connection->isConnected()) {
             return;
         }
--- a/Collector/MigrationsCollector.php
+++ b/Collector/MigrationsCollector.php
@@ -28,7 +28,7 @@ class MigrationsCollector extends DataCollector
         $this->flattener         = $migrationsFlattener;
     }
 
-    public function collect(Request $request, Response $response, ?Throwable $exception = null): void
+    public function collect(Request $request, Response $response, ?Throwable $exception = null)
     {
         if (! empty($this->data)) {
             return;
--- a/Collector/MigrationsCollector.php
+++ b/Collector/MigrationsCollector.php
@@ -10,6 +10,10 @@ use Doctrine\Migrations\Metadata\Storage\TableMetadataStorageConfiguration;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\DataCollector\DataCollector;
+use Throwable;
+
+use function count;
+use function get_class;
 
 class MigrationsCollector extends DataCollector
 {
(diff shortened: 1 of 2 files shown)


Collector/MigrationsCollector.php
```diff
@@ -21,17 +25,22 @@ class MigrationsCollector extends DataCollector
     public function __construct(DependencyFactory $dependencyFactory, MigrationsFlattener $migrationsFlattener)
     {
         $this->dependencyFactory = $dependencyFactory;
-        $this->flattener = $migrationsFlattener;
+        $this->flattener         = $migrationsFlattener;
     }
 
-    public function collect(Request $request, Response $response, \Throwable $exception = null)
+    public function collect(Request $request, Response $response, ?Throwable $exception = null): void
```
derrabus: Adding a return type is a BC break. We cannot do this in a bugfix release.

derrabus (changes requested):
Can you add a test that reproduces the problem you'd like to fix?

Collector/MigrationsCollector.php
```diff
@@ -21,17 +25,22 @@ class MigrationsCollector extends DataCollector
     public function __construct(DependencyFactory $dependencyFactory, MigrationsFlattener $migrationsFlattener)
     {
         $this->dependencyFactory = $dependencyFactory;
-        $this->flattener = $migrationsFlattener;
+        $this->flattener         = $migrationsFlattener;
     }
 
-    public function collect(Request $request, Response $response, \Throwable $exception = null)
+    public function collect(Request $request, Response $response, ?Throwable $exception = null)
```
oleg-andreyev, line 31: Please note that running phpcbf is adding return type.

oleg-andreyev:
> Can you add a test that reproduces the problem you'd like to fix?

It's hard to reproduce this scenario in CI, as we need to boot entire application (without configured dbal connection) and I haven't found any "functional tests" for this bundle.

oleg-andreyev:
@derrabus any chance that you'll be able to take a look at this?

nedelyux:
@derrabus Whats problem?

derrabus:
I'm sorry that nobody has picked up this PR after my initial triage. I still don't fully understand why problem is solved here. The collector needs a valid database connection. And without one, this whole bundle doesn't make much sense.

oleg-andreyev:
> I'm sorry that nobody has picked up this PR after my initial triage. I still don't fully understand why problem is solved here. The collector needs a valid database connection. And without one, this whole bundle doesn't make much sense.

It's been awhile, let me remember.

> Collector still requires an configure connection.
When starting new project and database is not yet configured, MigrationsCollector is trying to connect to database and because of PDO timeout, page feels are hanged.

IIRC initial problem is/was with clean project, when **database is not yet configured**.
So when I install this bundle it's automatically, it's automatically starts collecting information about executed migrations 
image
and since **database is not yet configured** it's trying to connect to a database (hosts does not exist) and it's taking 30s (default timeout) before it's throwing an exception.

and it's not very obvious that something is wrong and developers start digging into a problem why clean project is taking 30s to load.

At least this is what I remember.

oleg-andreyev:
^^^ and the idea was: avoid collecting migrations if connection is not configured.

oleg-andreyev:
and PR that was trying to fix similar issue: https://github.com/doctrine/DoctrineMigrationsBundle/pull/428 just wrapped code with try/catch which does not prevent of trying to make a connection

greg0ire:
This got closed when I deleted the 3.3.x branch. Sorry!

PR #491, opened 2023-03-14
Add type declaration to `MigrationCollector::reset()`

asuri0n:
Added the explicit @return type in `MigrationCollector::reset()` to suspend the deprecation info msg

image


Collector/MigrationsCollector.php
```diff
@@ -82,6 +82,9 @@ public function getData()
         return $this->data;
     }
 
+    /**
+     * {@inheritDoc}
```
derrabus: That's not enough to silence the deprecation emitted by Symfony's error handler. You need to document the return type explicitly.

```suggestion
     * @return void
```

SenseException (changes requested):
Please use a meaningful commit message as described in our [contribution guide](https://www.doctrine-project.org/contribute/index.html#crafting-meaningful-commit-messages).

asuri0n:
@SenseException I don't know how to edit my previous commit messages

stof:
you need to amend your git history locally and to force-push your branch after that

SenseException:
@asuri0n You can find infos on how to change the last commit here: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#_git_amend and do a force-push afterwards.

asuri0n:
well ... I think I screwed up the git history.... I will make a new merge request with clean commit message.

greg0ire:
> I will make a new merge request with clean commit message.

You don't need to, you can just force push once the history is good.

What you could do is

```
git reset 9432bc37ab3bc4f074425c669d8a63cc388b17df
git add --patch
git commit --verbose
git push --force-with-lease
```

greg0ire:
Cc @mpdude, regarding https://github.com/doctrine/doctrine-website/pull/473#discussion_r1129076152 :)

greg0ire:
@asuri0n there are 9 commits now. Have you tried what I recommended above?

mpdude:
If we‘d ask contributors to provide meaningful summaries in the opening comment, it would help everybody coming here for a review, and less experienced Git users could more easily change the text than with Git history rewrites and force-pushes.

Then, in the end, squash-merge and use the initial comment as the squash merge message. That also makes meaningless commits in the course of the PR („Fix CS“) irrelevant.

Would have saved us all some headache here, non? But that’s just my $.02, off topic here so I’ll just shut up 😉

greg0ire:
No no, you're right, I think it's just we didn't think of that! @asuri0n , can you please edit the body of the pull request, as suggested?

asuri0n:
It's a pretty simple pull request, I've updated the body, is it enough ?

greg0ire:
> to suspend the deprecation info msg

It would be nice to quote the message. Also, it would be nice to know why you fixed `reset()` but not `getData()`. Is there no message for that one?

I think it's a deprecation coming from Symfony, right? And I think maybe you considered using `@{inheritDoc}`, but that is not enough for some reason I don't remember?

derrabus:
> And I think maybe you considered using `@{inheritDoc}`, but that is not enough for some reason I don't remember?

https://github.com/doctrine/DoctrineMigrationsBundle/pull/491#discussion_r1135220070

greg0ire:
Oh right :facepalm:

derrabus (approved):

derrabus:
Thank you @asuri0n!

PR #492, opened 2023-03-28
hide deprecation warning

adeptofvoltron:
with symfony 6.1 I got such warning:

```
2023-03-28T12:23:12+02:00 [info] User Deprecated: Method "Symfony\Component\HttpKernel\Bundle\Bundle::build()" might add "void" as a native return type declaration in the future. Do the same in child class "Doctrine\Bundle\MigrationsBundle\DoctrineMigrationsBundle" now to avoid errors or add an explicit @return annotation to suppress this message.
```

such change repair it

### DIFF
--- a/DoctrineMigrationsBundle.php
+++ b/DoctrineMigrationsBundle.php
@@ -14,7 +14,7 @@ use Symfony\Component\HttpKernel\Bundle\Bundle;
  */
 class DoctrineMigrationsBundle extends Bundle
 {
-    public function build(ContainerBuilder $container)
+    public function build(ContainerBuilder $container): void
     {
       $container->addCompilerPass(new ConfigureDependencyFactoryPass());
     }


DoctrineMigrationsBundle.php
```diff
@@ -14,7 +14,7 @@
  */
 class DoctrineMigrationsBundle extends Bundle
 {
-    public function build(ContainerBuilder $container)
+    public function build(ContainerBuilder $container): void
```
greg0ire, line 17: The class is not final, so this is technically a breaking change. You should use a phpdoc annotation instead:

```suggestion
    /** @return void */
    public function build(ContainerBuilder $container)
```

stof: I also suggest adding `@final since 3.3.0` (or `3.4.0` depending on what is the next minor version) on that class

greg0ire: The next minor version should be 3.3.0.

derrabus (changes requested):
As @greg0ire said, we cannot add a return type to a non-final method in a minor release.

nicolas-grekas:
@adeptofvoltron up to do the suggested changes?

mickverm:
https://github.com/doctrine/DoctrineMigrationsBundle/pull/494 has been merged.

derrabus:
Thanks. Let's close this one then.

PR #516, opened 2023-11-13
Test against dev deps in CI

derrabus:
~~Blocked by #517~~

### DIFF
--- a/Tests/DependencyInjection/DoctrineCommandsTest.php
+++ b/Tests/DependencyInjection/DoctrineCommandsTest.php
@@ -24,6 +24,7 @@ use Doctrine\Migrations\Tools\Console\Command\VersionCommand;
 use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
 use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Component\Console\Command\LazyCommand;
 use Symfony\Component\Console\DependencyInjection\AddConsoleCommandPass;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\ParameterBag\ParameterBag;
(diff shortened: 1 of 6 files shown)


Tests/Fixtures/Migrations/Migration001.php
```diff
@@ -6,26 +6,11 @@
 
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\Migrations\AbstractMigration;
-use Symfony\Component\DependencyInjection\ContainerAwareInterface;
-use Symfony\Component\DependencyInjection\ContainerInterface;
 
-class Migration001 extends AbstractMigration implements ContainerAwareInterface
```
derrabus, line 12: The `ContainerAwareInterface` is going away in Symfony 7. My PR simply makes the tests pass, but we probably have to think about a proper replacement for this feature.

Tests/DependencyInjection/DoctrineCommandsTest.php
```diff
@@ -40,9 +41,12 @@ class DoctrineCommandsTest extends TestCase
      */
     public function testCommandRegistered(string $name, string $instance): void
     {
-        $application = $this->getApplication();
+        $actualCommand = $this->getApplication()->find($name);
```
greg0ire: This is a bit confusing… at this point `$actualCommand` might not in fact be the actual command. It will eventually be thanks to line 46 but right now it isn't. I think I would just use `$command` here.

Tests/Fixtures/Migrations/ContainerAwareMigration.php
```diff
@@ -0,0 +1,31 @@
...
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+class ContainerAwareMigration extends AbstractMigration implements ContainerAwareInterface
+{
+    /** @var ContainerInterface */
+    private $container;
+
+    public function up(Schema $schema): void
+    {
+        // TODO: Implement up() method.
```
greg0ire: I think we can remove this comment.

Tests/DependencyInjection/DoctrineCommandsTest.php
derrabus: Very well.

greg0ire (approved):

PR #518, opened 2023-11-13
Deprecate ContainerAwareMigrationFactory

derrabus:
`ContainerAwareInterface` will be removed in Symfony 7 which renders `ContainerAwareMigrationFactory` useless. I'm therefore deprecating this feature.

I realize that we don't have a replacement (yet), but since we already know that we can't keep this feature, we might as well deprecate it now. Someone who uses that feature might want to step forward and build a replacement.

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -3,7 +3,7 @@
     errorLevel="4"
     phpVersion="8.2"
     findUnusedBaselineEntry="true"
-    findUnusedCode="true"
+    findUnusedCode="false"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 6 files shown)


psalm.xml
```diff
@@ -3,7 +3,7 @@
     errorLevel="4"
     phpVersion="8.2"
     findUnusedBaselineEntry="true"
-    findUnusedCode="true"
+    findUnusedCode="false"
```
derrabus, line 6: For some reason, Psalm starts to detect classes used by `DoctrineMigrationsExtensionTest` as "unused" after my change. My motivation to investigate that is rather limited, I must admit. 😓

greg0ire (approved):

PR #523, opened 2024-01-03
Update config example to match flex recipe.

yivi:
Update config example to match the flex recipe, since the included example ("App\Migrations") went directly against it.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -41,12 +41,11 @@ application:
 
     # config/packages/doctrine_migrations.yaml
 
-    doctrine_migrations:
         # List of namespace/path pairs to search for migrations, at least one required
         migrations_paths:
-            'App\Migrations': '%kernel.project_dir%/src/App'
-            'AnotherApp\Migrations': '/path/to/other/migrations'
-            'SomeBundle\Migrations': '@SomeBundle/Migrations'
+            # namespace is arbitrary but should be different from App\Migrations
+            # as migrations classes should NOT be autoloaded
+            'DoctrineMigrations': '%kernel.project_dir%/migrations'
 
         # List of additional migration classes to be loaded, optional
         migrations:


greg0ire:
Can you please post a link to the PR that changed the flex recipe?

yivi:
I'll look for it. 

I'm only aware because of this because today cloned a new project today, and realized the docs didn't match the recipe advice.

yivi:
I don't know if I'm missing something, but it seems to have been like this forever?

https://github.com/symfony/recipes/commits/main/doctrine/doctrine-migrations-bundle/3.1/config/packages/doctrine_migrations.yaml

https://github.com/symfony/recipes/blob/main/doctrine/doctrine-migrations-bundle/2.2/config/packages/doctrine_migrations.yaml

https://github.com/symfony/recipes/blob/main/doctrine/doctrine-migrations-bundle/1.2/config/packages/doctrine_migrations.yaml

Resources/doc/index.rst
```diff
@@ -41,12 +41,11 @@ application:
...
     # config/packages/doctrine_migrations.yaml
 
-    doctrine_migrations:
         # List of namespace/path pairs to search for migrations, at least one required
         migrations_paths:
-            'App\Migrations': '%kernel.project_dir%/src/App'
-            'AnotherApp\Migrations': '/path/to/other/migrations'
-            'SomeBundle\Migrations': '@SomeBundle/Migrations'
+            # namespace is arbitrary but should be different from App\Migrations
+            # as migrations classes should NOT be autoloaded
```
greg0ire, line 47: Can you explain what this means?

yivi: As I understand it, that migrations should not be autoloaded by the Symfony service container build process.

greg0ire: Hmmm isn't what you're thinking of called auto discovery, or do you really mean PHP autoloading ?

yivi: No, I didn't mean autoloading in the sense of "class autoloading", as provided by the composer autoloader. I meant being loaded and configured by the dependency injection component.

In any case, not using an "App\" namespace would likely prevent both things in most scenarios targeted by the docs, so not sure if the difference is of actual relevance.

But frankly, I just meant to duplicate what the recipe said, since the docs seemed to me to be deferring to the recipe; and only including a default config for cases where Flex wasn't installed, and the recipe wasn't applied.

I actually didn't really think through if the recipe was actually right and the docs wrong, or vice versa. Maybe it's better I close my PR and let someone more knowledgeable or with more time to research to help with this.

Sorry I can't be of more help.

greg0ire: I was asking just in case you knew, or had the time to research, but I think I can find out. 

Following the last link you provided, a few extra clicks lead to https://github.com/symfony/recipes/pull/172

A bit more clicking leads to https://github.com/symfony/recipes/issues/343#issuecomment-721705155, which corroborates that it really is PHP autoloading that is meant here. I'm not saying I understand why autoloading a migration class would break the mechanism that loads migration classes from a directory, but I think PHP autoloading is really what was meant here. Maybe @stof can help with this?

Resources/doc/index.rst
```diff
@@ -41,12 +41,11 @@ application:
 
     # config/packages/doctrine_migrations.yaml
 
-    doctrine_migrations:
```
greg0ire, line 44: :thinking: why are you removing this? It still appears here: https://github.com/symfony/recipes/blob/0d06144589aeb8eff11025d398f724f9545fab2c/doctrine/doctrine-migrations-bundle/3.1/config/packages/doctrine_migrations.yaml#L1

greg0ire:
This got closed when I deleted the 3.4.x branch. Sorry!

Issue #521, opened 2023-12-14
Migration dependency injection in Symfony 7.0


TomBrouws:
The current documentation (https://symfony.com/bundles/DoctrineMigrationsBundle/current/index.html#migration-dependencies) refers to using `ContainerAwareInterface` in order to inject the entire container into a migration. In Symfony 6.4 this interface is deprecated, and it is removed in 7.0. (https://github.com/symfony/symfony-docs/issues/18440)

Is there a recommended way to tackle dependency injection for migrations in Symfony 7.0? Are code changes needed in this library in order to replicate the old functionality, or is an update of the docs sufficient?

derrabus:
> Is there a recommended way to tackle dependency injection for migrations in Symfony 7.0?

No. As you said, the upstream interface is gone. We don't have a replacement (yet?).

> Are code changes needed in this library in order to replicate the old functionality,

Yes.

> or is an update of the docs sufficient?

The docs change would currently be to remove the mention of it from the docs or at least add a note that the feature is gone when upgrading to Symfony 7.

derrabus:
That being said, injecting the whole container is highly discouraged by Symfony. If we were to build a replacement, we should either:

* Make it possible (not mandatory!) to register migrations as services
* Inject a service locator instead, e.g. by allowing migrations to be service subscribers.

TomBrouws:
Thanks for your reply. Good to know there's no replacement yet.

> injecting the whole container is highly discouraged by Symfony

Yes, by 'replicate' I meant in a way that is idiomatic in the newest Symfony version, so your suggestions make sense.

demcy:
Please, give an working example of custom migration factories that should be used to inject additional dependencies into migrations.

derrabus:
> Please, give an working example of custom migration factories that should be used to inject additional dependencies into migrations.

Nobody can give you a working example for a feature that doesn't exist.

hugo-fasone:
I'm somewhat perplexed by the changes since Symfony 7.0. Since the ContainerAwareInterface is no longer available, does this mean there's no way to inject or access services within migrations?

As a newcomer to Symfony, my understanding is that the framework heavily relies on "services" and dependency injection. However, it seems there's no way to utilize these services within migrations since they aren't registered as services. This limitation might be a significant drawback for adopting Symfony 7.0 in my scenario.

I've noticed that this issue hasn't generated much discussion or feedback, which surprises me given its potential impact.

[...]

PR #524, opened 2024-01-21
Add service migrations (POC)

HypeMC:
Inspired by https://github.com/doctrine/DoctrineMigrationsBundle/issues/521#issuecomment-1855593824

This is a proposed implementation for registering migrations as services.

To get this working Symfony needs to be able to autoload migrations:

```json
    "autoload": {
        "psr-4": {
            "App\\": "src/",
            "DoctrineMigrations\\": "migrations/"
        }
    },
```

Then, import the migrations as services:

```yaml
    DoctrineMigrations\:
        resource: '../migrations/'
```

Of course, an alternative approach would be to move the `migrations` folder to `src` and change the namespace to `App\Migrations`.

Once that is done, all you need to do is override the constructor and inject the desired services.

```php
final class Version20240121015756 extends AbstractMigration
{
    public function __construct(
        Connection              $connection,
        LoggerInterface         $logger,
        private RouterInterface $router,
    ) {
        parent::__construct($connection, $logger);
    }

    public function up(Schema $schema): void
    {
        dump($this->router);

        // ...
    }

    // ...
}
```

I haven't added any tests yet, as I was hoping to get some feedback on my approach first.

### DIFF
--- a/config/services.xml
+++ b/config/services.xml
@@ -11,6 +11,14 @@
             <argument type="service" id="doctrine.migrations.configuration_loader"/>
             <argument></argument>
             <argument type="service" id="logger" on-invalid="null"></argument>
+            <call method="setDefinition">
+                <argument>Doctrine\Migrations\Finder\MigrationFinder</argument>
+                <argument type="service_closure" id="doctrine.migrations.migration_finder" />
+            </call>
+            <call method="setDefinition">
+                <argument>Doctrine\Migrations\MigrationsRepository</argument>
+                <argument type="service_closure" id="doctrine.migrations.migrations_repository" />
+            </call>
         </service>
 
         <service id="doctrine.migrations.configuration_loader" class="Doctrine\Migrations\Configuration\Migration\ExistingConfiguration" public="false">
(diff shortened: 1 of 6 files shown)


connorhu:
Here you can find a similar but much simpler implementation. My solution works without extending the autoloader.
https://github.com/doctrine/DoctrineMigrationsBundle/compare/3.3.x...connorhu:DoctrineMigrationsBundle:3.3.x

derrabus:
This PR will require to register migrations as services, right? If we go down that road, why would I need them to be service subscribers? I mean, I could just inject all dependencies directly, couldn't I?

HypeMC:
@derrabus The reason for my approach are the standard migration arguments: `Connection $connection, LoggerInterface $logger`. `Connection` would get autowired to the connection in Doctrine bundle, not the result of `doctrine.migrations.dependency_factory::getConnection()`, same goes for the logger. Perhaps I could use bindings instead :thinking: . I'll give it a try.

HypeMC:
@derrabus I've updated the implementation to use bound arguments instead of service locators.

AlexOstrovsky:
@greg0ire any plan when this will be merged?

greg0ire:
No

AlexOstrovsky:
@greg0ire what is blocking this from beeing merged?

greg0ire:
For starters:

- [x] There are conflicts.
- [x] The CI should be green.
- [ ] The conversation with @derrabus must be resolved.

HypeMC:
@greg0ire There doesn't seem to be a branch 3.6.x.

greg0ire:
Oh right sorry, I got confused with another project (`DoctrineFixturesBundle`). 3.4.x is correct.

greg0ire:
I know it's just a POC for now, but since I'm asked what's blocking, I'd add that docs are probably necessary for this sort of thing.

connorhu:
I started testing this PR with real world code. You should work on making this unnecessary:
```
    DoctrineMigrations\:
        resource: '../migrations/'
```

greg0ire:
@HypeMC should we close this?

greg0ire:
This got closed when I deleted the 3.4.x branch. Sorry!

HypeMC:
> @HypeMC should we close this?

@greg0ire I'm still up for working on this, but I'd appreciate some feedback to confirm whether there's any interest from your side. I still believe this would be a useful feature to have.

greg0ire:
Given how much activity there is in https://github.com/doctrine/DoctrineMigrationsBundle/issues/521 I'd say yes, there is interest. I don't use migrations myself so it's hard for me to judge.

PR #527, opened 2024-03-14
Exclude file from static analysis

greg0ire:
That file contains a class that implements a non-existing interface
(when using Symfony 7+): `ContainerAwareInterface`.


SenseException (approved):
This PR is about the current failing static analysis of a test class. This should be fine for the moment.

PR #530, opened 2024-03-22
Introduce EnvironmentAwareMigrationInterface

alexndlm:
In migrations, it's best to avoid using dynamic values or services that could cause unexpected issues. However, including the current environment can be really helpful. It lets us skip migrations for testing or production and set different default values depending on the environment.


alexndlm:
@derrabus wdyt?

greg0ire:
It seems there are CI jobs failing. Please take a look at [this guide](https://www.doctrine-project.org/contribute/#dealing-with-checks-and-tools) for more on how to handle those.

alexndlm:
@greg0ire, thanks, fixed.

greg0ire:
Blocked by #531

greg0ire:
Retargeting to 3.4.x since this is no bugfix

ERuban:
Any news on that?

derrabus:
I'd like to close this PR. I don't think, adding more FooAwareMigrationInterfaces are the proper way forward. Instead, we should find a way to leverage DI for migrations as I layed out in https://github.com/doctrine/DoctrineMigrationsBundle/issues/521#issuecomment-1855593824.

PR #531, opened 2024-03-25
Address deprecation from DoctrineBundle

greg0ire:
This fixes the build, which is sensitive to deprecations on this repository.

### DIFF
--- a/Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
+++ b/Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Doctrine\Bundle\MigrationsBundle\Tests\DependencyInjection;
 
+use Composer\InstalledVersions;
+use Composer\Semver\VersionParser;
 use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\CacheCompatibilityPass;
 use Doctrine\Bundle\DoctrineBundle\DependencyInjection\DoctrineExtension;
 use Doctrine\Bundle\DoctrineBundle\Registry;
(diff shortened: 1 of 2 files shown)


stof:
Can't we always set this configuration entry ? Or do we still support old versions of the bundle that don't have this setting ?

greg0ire:
> Can't we always set this configuration entry ? Or do we still support old versions of the bundle that don't have this setting ?

That's what I did in [my initial try](https://github.com/doctrine/DoctrineMigrationsBundle/commit/55d309aee66d4fe13f1c4e833655de8194b35123), and we support old versions of PHP which prevent installing a recent enough version of the bundle (or at least that's what I think is happening).

Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
```diff
@@ -285,6 +287,10 @@ public function testPrefersEntityManagerOverConnection(): void
             'migrations_paths' => ['DoctrineMigrationsTest' => 'a'],
         ];
         $ormConfig = trait_exists(LazyGhostTrait::class) ? ['enable_lazy_ghost_objects' => true] : [];
+        if (PHP_VERSION_ID >= 73000) {
+            $ormConfig['controller_resolver'] = ['auto_mapping' => false];
+        }
```
derrabus: This looks like it would break if I did a low-deps run on PHP 7.3. 🤔

greg0ire: It would, I couldn't find a reliable way to detect the right version of the bundle. Since this detection code only occurs in tests, I thought it would not be a big deal.

derrabus: Shall we try `Composer\InstalledVersions` instead then?

Tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
```diff
@@ -285,6 +287,10 @@ public function testPrefersEntityManagerOverConnection(): void
             'migrations_paths' => ['DoctrineMigrationsTest' => 'a'],
         ];
         $ormConfig = trait_exists(LazyGhostTrait::class) ? ['enable_lazy_ghost_objects' => true] : [];
+        if (InstalledVersions::satisfies(new VersionParser(), 'doctrine/doctrine-bundle', '^2.7.1 ')) {
```
greg0ire, line 290: I'm using the version at which the option was introduced, not the version at which it was deprecated. See https://github.com/doctrine/DoctrineBundle/commit/71d64113156dd7f13eebd3dd1147f943da77bb40

derrabus (approved):

stof (approved):

PR #533, opened 2024-03-26
Cover all directories with our tooling

greg0ire:
The goal is to hold new code with higher standards, and to have the same configuration on all branches. Since I have no proof addressing static analysis issues fixes any actual bugs, let us ignore them all on this branch.

I initially wanted to modernize the directory structure just like in https://github.com/doctrine/DoctrineBundle/pull/1767, but this needs to be done first.


derrabus (approved):

derrabus:
For 3.4?

greg0ire:
> The goal is to hold new code with higher standards, and to have the same configuration on all branches.

If we target 3.4 there is a risk we will have to tweak stuff related to CS or SA while merging up, so I'd rather we stick with 3.3. I ignored SA errors so that this is not controversial, so all the changes that happen in `src` are style/docblock changes.

PR #535, opened 2024-03-26
Switch to a modern directory structure

greg0ire:
Here are the steps I followed (probably not an exhaustive list, but can be useful to migrate more bundles):

- `mkdir src`
- `git mv Collector DependencyInjection DoctrineMigrationsBundle.php MigrationsFactory src`
- adjust autoloading path
- tweak PHPUnit config
- tweak Psalm config
- tweak phpcs config
- `git mv Resources/config config`
- `git mv Resources/docs doc`
- overrode `Bundle::getPath()`
- `git mv Tests tests`
- `git mv Resources/views templates`


derrabus (approved):

connorhu:
@greg0ire +1 item to the todo list: update .gitattributes
I will do an archive from 3.4.x and check out what needs to be updated and I'll open a PR.

greg0ire:
Ah indeed, I missed at least `Tests`

PR #540, opened 2024-04-08
Fix .gitattribute 

connorhu:
Mostly fixes because of #535
- updated Tests
- removed phpstan neon
- added phpstan baseline
- added docs

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,13 +1,14 @@
 * text=auto
 
-/Tests export-ignore
+/docs export-ignore
+/tests export-ignore
 .doctrine-project.json export-ignore
 .gitattributes export-ignore
 /.github export-ignore
 .gitignore export-ignore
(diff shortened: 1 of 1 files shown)


connorhu:
A couple of weeks ago I was playing around with the problem of whether there is any way to check the export-ignore setting with "deny all".
I ended up with a GHA task, but I'm not completely satisfied with it yet (I'm not a git wizard).

In a file it was listed what was allowed in the root directory and in case of PR a job checked if the ones in the root were allowed.

SenseException:
Thanks for the contribution. Could you please [squash](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#_squashing) your commits?

connorhu:
@SenseException Sure!

greg0ire:
Thanks @connorhu !

PR #571, opened 2025-01-12
Drop unsupported versions of PHP

greg0ire:
This should allow us to do many breaking changes related to new features of PHP.

I'm buping to 8.1 and not higher for the sake of consistency with other recent versions of Doctrine packages. Do let me know if you think I should bump higher.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         }
     ],
     "require": {
-        "php": "^7.2|^8.0",
+        "php": "^8.1",
         "symfony/deprecation-contracts": "^2.1 || ^3",
         "symfony/framework-bundle": "^5.4 || ^6.0 || ^7.0",
         "doctrine/doctrine-bundle": "^2.4",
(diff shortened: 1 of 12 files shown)


composer.json
```diff
@@ -37,8 +37,9 @@
         "phpstan/phpstan-symfony": "^1.3",
         "doctrine/orm": "^2.6 || ^3",
         "doctrine/persistence": "^2.0 || ^3 ",
-        "symfony/phpunit-bridge": "^6.3 || ^7",
-        "symfony/var-exporter": "^5.4 || ^6 || ^7"
+        "symfony/console": "^6.4.17",
```
greg0ire: Had to add this constraint because of an indirect deprecation that caused the prefer-lowest job to fail.

derrabus: So we're not testing with Console 7? That would be really bad.

greg0ire: Oh yeah sorry, I will add it.

derrabus (approved):
I've merged up your changes on 3.3.x. If you want, you can rebase and throw PHPStan 1 our of composer.json.

derrabus (approved):

PR #575, opened 2025-01-12
Upgrade to PHPUnit 11

greg0ire:
The use of listeners is no longer possible, but we can replace the deprecation detection functionality with native PHPUnit features.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
     },
     "require-dev": {
         "composer/semver": "^3.0",
-        "phpunit/phpunit": "^9.6.22",
+        "phpunit/phpunit": "^10.5.40 || ^11.5",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -28,7 +28,7 @@
     },
     "require-dev": {
         "composer/semver": "^3.0",
-        "phpunit/phpunit": "^9.6.22",
+        "phpunit/phpunit": "^10.5 || ^11.5",
```
derrabus:
```suggestion
        "phpunit/phpunit": "^10.5.40 || ^11.5",
```

Let's make sure, the low deps run is run on a fairly recent version of PHPUnit.

derrabus (approved):

PR #587, opened 2025-01-19
Apply schema filter more selectively

### LINKED ISSUE #584: After v3.4 all migrations shown as not executed in dev toolbar
### Bug Report

Just checked out the new release and all of my past migrations shown as not executed:

Image

| Version                                     | 3.4.0
| Previous Version if the bug is a regression | 3.3.1

#### Summary

cache cleared, metadatastorage synced, migration table contains all 158 entry. The toolbar still shows everything as new

### LINKED ISSUE #585: Regression: Migrations table already exists error in 3.4
### Bug Report

| Version                                     | 3.4.0
| Previous Version if the bug is a regression | 3.3.1

#### Summary

After upgrading from 3.3.1 to 3.4.0 our migrations fail with

> [Doctrine\DBAL\Exception\TableExistsException (1050)]
  An exception occurred while executing a query: SQLSTATE[42S01]: Base table or view already exists: 1050 Table 'doctrine_migration_versions' already exists

<details>
<summary>Full output and trace</summary>

```
15:06:22 NOTICE    [app] Running Doctrine migrations ...
15:06:22 CRITICAL  [console] Error thrown while running command "-vv hc:install". Message: "An exception occurred while executing a query: SQLSTATE[42S01]: Base table or view already exists: 1050 Table 'doctrine_migration_versions' already exists" ["exception" => Doctrine\DBAL\Exception\TableExistsException^ { …},"command" => "-vv hc:install","message" => "An exception occurred while executing a query: SQLSTATE[42S01]: Base table or view already exists: 1050 Table 'doctrine_migration_versions' already exists"]

In ExceptionConverter.php line 38:

  [Doctrine\DBAL\Exception\TableExistsException (1050)]
  An exception occurred while executing a query: SQLSTATE[42S01]: Base table or view already exists: 1050 Table 'doctrine_migration_versions' already exists


Exception trace:
  at /var/www/html/vendor/doctrine/dbal/src/Driver/API/MySQL/ExceptionConverter.php:38
 Doctrine\DBAL\Driver\API\MySQL\ExceptionConverter->convert() at /var/www/html/vendor/doctrine/dbal/src/Connection.php:1414
 Doctrine\DBAL\Connection->handleDriverException() at /var/www/html/vendor/doctrine/dbal/src/Connection.php:1356
 Doctrine\DBAL\Connection->convertExceptionDuringQuery() at /var/www/html/vendor/doctrine/dbal/src/Connection.php:888
 Doctrine\DBAL\Connection->executeStatement() at /var/www/html/packages/doctrine/src/DBAL/Con.php:55
 Octaved\Doctrine\DBAL\Con->executeStatement() at /var/www/html/vendor/doctrine/dbal/src/Schema/AbstractSchemaManager.php:779
 Doctrine\DBAL\Schema\AbstractSchemaManager->executeStatements() at /var/www/html/vendor/doctrine/dbal/src/Schema/AbstractSchemaManager.php:514
 Doctrine\DBAL\Schema\AbstractSchemaManager->createTable() at /var/www/html/vendor/doctrine/migrations/src/Metadata/Storage/TableMetadataStorage.php:172
 Doctrine\Migrations\Metadata\Storage\TableMetadataStorage->ensureInitialized() at /var/www/html/vendor/doctrine/migrations/src/Tools/Console/Command/MigrateCommand.php:158
 Doctrine\Migrations\Tools\Console\Command\MigrateCommand->execute() at /var/www/html/vendor/symfony/console/Command/Command.php:279
 Symfony\Component\Console\Command\Command->run() at /var/www/html/vendor/symfony/console/Command/LazyCommand.php:77
 Symfony\Component\Console\Command\LazyCommand->run() at /var/www/html/packages/core/src/HContent/CoreBundle/Command/InstallCommand.php:67
 HContent\CoreBundle\Command\InstallCommand->runDoctrineMigrations() at /var/www/html/packages/core/src/HContent/CoreBundle/Command/InstallCommand.php:43
 HContent\CoreBundle\Command\InstallCommand->execute() at /var/www/html/vendor/symfony/console/Command/Command.php:279
 Symfony\Component\Console\Command\Command->run() at /var/www/html/vendor/symfony/console/Application.php:1094
 Symfony\Component\Console\Application->doRunCommand() at /var/www/html/vendor/symfony/framework-bundle/Console/Application.php:123
 Symfony\Bundle\FrameworkBundle\Console\Application->doRunCommand() at /var/www/html/vendor/symfony/console/Application.php:342
 Symfony\Component\Console\Application->doRun() at /var/www/html/vendor/symfony/framework-bundle/Console/Application.php:77
 Symfony\Bundle\FrameworkBundle\Console\Application->doRun() at /var/www/html/vendor/symfony/console/Application.php:193
 Symfony\Component\Console\Application->run() at /var/www/html/vendor/symfony/runtime/Runner/Symfony/ConsoleApplicationRunner.php:49
 Symfony\Component\Runtime\Runner\Symfony\ConsoleApplicationRunner->run() at /var/www/html/vendor/autoload_runtime.php:29
 require_once() at /var/www/html/app/api/bin/console:11
```

</details>

I could not find any upgrade notes that would hint this could happen in this minor update. Nor could I see anything obvious in the code changes between 3.3.1 and 3.4.0.

Is this a bug or did we something wrong? 

We have a few middlewares registered, but they don't appear to be relevant here. And our custom SqlGenerator is not at play here, either.

#### Current behavior

Migrations table is tried to be created.

#### Expected behavior

Migrations table already exist and should not be created.

#### How to reproduce

<details>
<summary>Our config</summary>

```yml
doctrine_migrations:
  migrations_paths:
    'Octaved\Doctrine\MigrationVersions': '%hc.path.packages%/doctrine/migrations'

  storage:
    table_storage:
      table_name: 'doctrine_migration_versions'
      version_column_name: 'version'
      version_column_length: 191
      executed_at_column_name: 'executed_at'
      execution_time_column_name: 'execution_time'

  # Possible values: "BY_YEAR", "BY_YEAR_AND_MONTH", false
  organize_migrations: 'BY_YEAR' # this config is not used because we have overwritten the Generator

  # Run all migrations in a transaction.
  all_or_nothing: false

  # Adds an extra check in the generated migrations to ensure that is executed on the same database type.
  check_database_platform: false

  # Whether or not to wrap migrations in a single transaction.
  transactional: false
```

</details>

greg0ire:
It is more careful, since there seems to be a lot of situations where we do not want to apply the filter.
Let us take the opposite approach and:
- Have the filter disabled by default, instead of enabled by default.
- Enable it for precise commands where we know we need it.

This means the ORM is no longer just a dev dependency.

Fixes #584, fixes #585

# How can I test this?
```shell
composer config repositories.greg0ire vcs https://github.com/greg0ire/DoctrineMigrationsBundle
composer require doctrine/doctrine-migrations-bundle "dev-selective-filter as 3.4.0"
```

### DIFF
--- a/tests/Collector/EventListener/SchemaFilterListenerTest.php
+++ b/tests/Collector/EventListener/SchemaFilterListenerTest.php
@@ -7,6 +7,10 @@ namespace Doctrine\Bundle\MigrationsBundle\Tests\Collector\EventListener;
 use Doctrine\Bundle\MigrationsBundle\EventListener\SchemaFilterListener;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\Migrations\Tools\Console\Command\DoctrineCommand;
+use Doctrine\ORM\Tools\Console\Command\AbstractEntityManagerCommand;
+use Doctrine\ORM\Tools\Console\Command\SchemaTool\UpdateCommand;
+use Doctrine\ORM\Tools\Console\Command\ValidateSchemaCommand;
+use Doctrine\ORM\Tools\Console\EntityManagerProvider;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\Console\Event\ConsoleCommandEvent;
 use Symfony\Component\Console\Input\ArrayInput;
@@ -14,18 +18,17 @@ use Symfony\Component\Console\Output\NullOutput;
 
 class SchemaFilterListenerTest extends TestCase
 {
-    public function testItFiltersOutMigrationMetadataTableByDefault(): void
+    public function testItFiltersNothingByDefault(): void
     {
         $listener = new SchemaFilterListener('doctrine_migration_versions');
-
-        self::assertFalse($listener(new Table('doctrine_migration_versions')));
+        self::assertTrue($listener(new Table('doctrine_migration_versions')));
         self::assertTrue($listener(new Table('some_other_table')));
     }
 
-    public function testItDisablesItselfWhenTheCurrentCommandIsAMigrationsCommand(): void
+    public function testItFiltersNothingWhenNotRunningSpecificCommands(): void
     {
         $listener          = new SchemaFilterListener('doctrine_migration_versions');
-        $migrationsCommand = new class extends DoctrineCommand {
+        $migrationsCommand = new class () extends DoctrineCommand {
         };
 
         $listener->onConsoleCommand(new ConsoleCommandEvent(
@@ -35,5 +38,33 @@ class SchemaFilterListenerTest extends TestCase
         ));
 
         self::assertTrue($listener(new Table('doctrine_migration_versions')));
+        self::assertTrue($listener(new Table('some_other_table')));
+    }
+
+    /**
+     * @param class-string<AbstractEntityManagerCommand> $command
+     *
+     * @dataProvider getCommands
+     */
+    public function testItFiltersOutMigrationMetadataTableWhenRunningSpecificCommands(string $command): void
+    {
+        $listener   = new SchemaFilterListener('doctrine_migration_versions');
+        $ormCommand = new $command($this->createStub(EntityManagerProvider::class));
+
+        $listener->onConsoleCommand(new ConsoleCommandEvent(
+            $ormCommand,
+            new ArrayInput([]),
+            new NullOutput()
+        ));
+
+        self::assertFalse($listener(new Table('doctrine_migration_versions')));
+        self::assertTrue($listener(new Table('some_other_table')));
+    }
+
+    /** @return iterable<string, array{class-string<AbstractEntityManagerCommand>}> */
+    public static function getCommands(): iterable
+    {
+        yield 'orm:validate-schema' => [ValidateSchemaCommand::class];
+        yield 'orm:schema-tool:update' => [UpdateCommand::class];
     }
 }
--- a/src/EventListener/SchemaFilterListener.php
+++ b/src/EventListener/SchemaFilterListener.php
@@ -40,11 +40,6 @@ final class SchemaFilterListener
         return $asset !== $this->configurationTableName;
     }
 
-    private function disable(): void
-    {
-        $this->enabled = false;
-    }
-
     public function onConsoleCommand(ConsoleCommandEvent $event): void
     {
         $command = $event->getCommand();
@@ -53,6 +48,6 @@ final class SchemaFilterListener
             return;
         }
 
-        $this->disable();
+        $this->enabled = false;
     }
 }
--- a/src/EventListener/SchemaFilterListener.php
+++ b/src/EventListener/SchemaFilterListener.php
@@ -5,7 +5,8 @@ declare(strict_types=1);
 namespace Doctrine\Bundle\MigrationsBundle\EventListener;
 
 use Doctrine\DBAL\Schema\AbstractAsset;
-use Doctrine\Migrations\Tools\Console\Command\DoctrineCommand;
+use Doctrine\ORM\Tools\Console\Command\SchemaTool\UpdateCommand;
+use Doctrine\ORM\Tools\Console\Command\ValidateSchemaCommand;
 use Symfony\Component\Console\Event\ConsoleCommandEvent;
 
 /**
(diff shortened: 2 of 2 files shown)


tests/Collector/EventListener/SchemaFilterListenerTest.php
```diff
@@ -6,26 +6,26 @@
...
         self::assertTrue($listener(new Table('some_other_table')));
     }
 
-    public function testItDisablesItselfWhenTheCurrentCommandIsAMigrationsCommand(): void
+    public function testItFiltersOutMigrationMetadataTableWhenRunningSpecificCommands(): void
     {
         $listener          = new SchemaFilterListener('doctrine_migration_versions');
-        $migrationsCommand = new class extends DoctrineCommand {
+        $migrationsCommand = new class ($this->createStub(EntityManagerProvider::class)) extends ValidateSchemaCommand {
         };
```
SenseException: Isn't this still a valid test as the opposite for `testItFiltersOutMigrationMetadataTableWhenRunningSpecificCommands()`?

VincentLanglet:
> This means the ORM is no longer just a dev dependency.

Isn't useless for mongo-odm users ?

Since you're doing only instanceof check, I'm not sure you need to require the package.
Can't we do `$foo instanceof SomeClass` even if the class doesn't exists ? (We can also add a class_exists check)

The best way to solve this might be to
- Introduce an interface somewhere
- Use it in orm bundle (ValidateCommand implements InterfaceFoo)
- Change the check here (if !$command instanceof InterfaceFoo)

But I didn't found a package used by both orm and migrationBundle. (That would end in the famous doctrine/contract package). Until them I think you can keep the instanceof check without requiring ORM.

greg0ire:
> Isn't useless for mongo-odm users ?

You're right, I'll revert this change.

tests/Collector/EventListener/SchemaFilterListenerTest.php
greg0ire: The name is no longer valid, since the filter no longer disables itself (it's disabled by default and stays so until enabled).

There is another test that tests that it's disabled by default.

SenseException: It isn't about the name but about the 2 cases where we have an instance of UpdateCommand or ValidateSchemaCommand and not having an instance of ValidateSchemaCommand or UpdateCommand. `onConsoleCommand` is currently only covering one of those cases. UpdateCommand isn't covered either.

victor-paumier:
For https://github.com/doctrine/DoctrineMigrationsBundle/issues/584, this solution currently hides the toolbar indicator. If we reactivate it, it still indicates that none of the migrations have been executed
Capture d’écran 2025-01-21 à 15 49 48


On the profiler itself, it corrects the problem well
Capture d’écran 2025-01-21 à 16 03 58

greg0ire:
@victor-paumier I don't understand how that's possible, given that with this solution, the filter will never be enabled in the context of HTTP. Maybe there is a cache you need to clear?

victor-paumier:
> @victor-paumier I don't understand how that's possible, given that with this solution, the filter will never be enabled in the context of HTTP. Maybe there is a cache you need to clear?

I just enable it manually, I just wanted to know if it was the expected result.
So does it means that this PR will deprecate the [toolbar block of the template](https://github.com/doctrine/DoctrineMigrationsBundle/blob/98725510d44e09fed5127b9dabf8bc463bf92686/templates/Collector/migrations.html.twig#L3-L59)? Or do you plan to fix it later?

greg0ire:
> I just enable it manually

You enabled the filter? Why?

tests/Collector/EventListener/SchemaFilterListenerTest.php
greg0ire: Addressed, I think.

greg0ire:
ping @derrabus @SenseException :pray:

SenseException (approved):
I didn't understand the use case of @victor-paumier though. Maybe it can be addressed separately when a deeper explanation is available.

greg0ire:
Me neither, and they don't answer my questions so :shrug:

PR #590, opened 2025-01-28
Add support for multiple connections managed by migrations

tobias-93:
References https://github.com/doctrine/migrations/issues/1406, which fixes the filtering of the migrations table for the default connection but not for any other connections.
This PR applies the filter to each of the Doctrine connections.
Not sure what else is required in terms of tests/documentation etc.

### DIFF
--- a/config/services.xml
+++ b/config/services.xml
@@ -148,7 +148,7 @@
 
         <service id="doctrine_migrations.schema_filter_listener" class="Doctrine\Bundle\MigrationsBundle\EventListener\SchemaFilterListener">
             <tag name="kernel.event_listener" event="console.command" method="onConsoleCommand" />
-<!--            <tag name="doctrine.dbal.schema_filter" /> This tag is dynamically added for each connection-->
+<!--        <tag name="doctrine.dbal.schema_filter" /> This tag is dynamically added for each connection -->
         </service>
 
     </services>
--- a/config/services.xml
+++ b/config/services.xml
@@ -148,7 +148,7 @@
 
         <service id="doctrine_migrations.schema_filter_listener" class="Doctrine\Bundle\MigrationsBundle\EventListener\SchemaFilterListener">
             <tag name="kernel.event_listener" event="console.command" method="onConsoleCommand" />
-            <tag name="doctrine.dbal.schema_filter" />
+<!--            <tag name="doctrine.dbal.schema_filter" /> This tag is dynamically added for each connection-->
         </service>
 
     </services>
--- a/src/DependencyInjection/DoctrineMigrationsExtension.php
+++ b/src/DependencyInjection/DoctrineMigrationsExtension.php
@@ -124,6 +124,15 @@ class DoctrineMigrationsExtension extends Extension
             }
 
             $configurationDefinition->addMethodCall('setMetadataStorageConfiguration', [new Reference('doctrine.migrations.storage.table_storage')]);
+
+            // Add tag to the filter for each Doctrine connection, so the table is ignored for multiple connections
+            if ($container->hasParameter('doctrine.connections')) {
+                /** @var array<string, string> $connections */
+                $connections = $container->getParameter('doctrine.connections');
+                foreach (array_keys($connections) as $connection) {
+                    $filterDefinition->addTag('doctrine.dbal.schema_filter', ['connection' => $connection]);
+                }
+            }
         }
 
         if ($config['em'] !== null && $config['connection'] !== null) {
--- a/tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
+++ b/tests/DependencyInjection/DoctrineMigrationsExtensionTest.php
@@ -279,6 +279,10 @@ class DoctrineMigrationsExtensionTest extends TestCase
         $di = $container->get('doctrine.migrations.dependency_factory');
         self::assertInstanceOf(DependencyFactory::class, $di);
         self::assertSame($doctrine->getConnection('custom'), $di->getConnection());
+        // Check if the multiple connections also added the required tags to the filter
+        $filterDefinition = $container->findDefinition('doctrine_migrations.schema_filter_listener');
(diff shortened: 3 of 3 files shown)


arderyp:
Is "my_custom_connection" on the example the name of a migration table? This example/naming confuses me

tobias-93:
> Is "my_custom_connection" on the example the name of a migration table? This example/naming confuses me

No, it is the name of a connection (see https://symfony.com/doc/current/reference/configuration/doctrine.html#doctrine-dbal-configuration). The name of the migrations table (of which there is only one per database/connection) is configured in the storage (see https://symfony.com/bundles/DoctrineMigrationsBundle/current/index.html#configuration)

src/DependencyInjection/Configuration.php
```diff
@@ -141,6 +141,11 @@ public function getConfigTreeBuilder(): TreeBuilder
                     ->info('Whether or not to wrap migrations in a single transaction.')
                     ->defaultTrue()
                 ->end()
+                ->arrayNode('connections')
+                    ->info('A list of connections managed by migrations, for which the migrations administrative table shall be filtered. The "default" connection is automatically filtered.')
+                    ->prototype('scalar')->end()
+                    ->defaultValue([])
+                ->end()
```
greg0ire: Having to re-specify connections here feels clumsy. Users won't understand why they need to do so.

src/DependencyInjection/DoctrineMigrationsExtension.php
```diff
@@ -135,6 +135,13 @@ public function load(array $configs, ContainerBuilder $container): void
         $container->setParameter('doctrine.migrations.preferred_em', $config['em']);
         $container->setParameter('doctrine.migrations.preferred_connection', $config['connection']);
 
+        if ($config['connections'] !== []) {
+            $filterDefinition = $container->getDefinition('doctrine_migrations.schema_filter_listener');
+            foreach ($config['connections'] as $connection) {
```
greg0ire: Rather than looping on this, how about creating a compiler pass that uses this parameter: https://github.com/doctrine/DoctrineMigrationsBundle/blob/e858ce0f5c12b266dce7dce24834448355155da7/src/DependencyInjection/CompilerPass/ConfigureDependencyFactoryPass.php#L82

arderyp:
> No, it is the name of a connection

How then does this relate to or fix the issue of the default dropping of the migrations table (the issues you posted this PR to as a solution)

Forgive my confusion, I must be missing something

src/DependencyInjection/DoctrineMigrationsExtension.php
tobias-93: You're right, it makes much more sense. I have simplified my fix.

tobias-93:
With the latest update, the configuration value is no longer needed. I have verified it and it works correctly for multiple connections now.

greg0ire (approved):

SenseException (changes requested):
This feature looks testable (tests/DependencyInjection/DoctrineMigrationsExtensionTest.php)

config/services.xml
```diff
@@ -148,7 +148,7 @@
 
         <service id="doctrine_migrations.schema_filter_listener" class="Doctrine\Bundle\MigrationsBundle\EventListener\SchemaFilterListener">
             <tag name="kernel.event_listener" event="console.command" method="onConsoleCommand" />
-            <tag name="doctrine.dbal.schema_filter" />
+<!--            <tag name="doctrine.dbal.schema_filter" /> This tag is dynamically added for each connection-->
```
greg0ire:
```suggestion
<!--        <tag name="doctrine.dbal.schema_filter" /> This tag is dynamically added for each connection -->
```

SenseException (approved):

arderyp:
> With the latest update, the configuration value is no longer needed. I have verified it and it works correctly for multiple connections now.

So... a return to the pre-break default behavior of assuming/expecting/ignoring a `migration_versions` table per connection? If not, would you mind explaining more?

tobias-93:
@arderyp indeed, the configured migrations table name is filtered in the migrations context. Pre-break this needed to be configured in the DBAL context if I'm not mistaken, which is now done automatically

arderyp:
Thanks @tobias-93 

I'm not entirely clear on your latest comment.  Here is how I remember the series of events

Years ago I had to configure doctrine migrations to ignore my migrations table. Then eventually DM changed something such that the migrations table that was configured was ignored by default.  Then, something broke, and required me to explicitly ignore the migrations tables again, despite them being explicitly configured (and thus eligible to discovering and ignoring).

Are you saying that this PR will make it so DM automatically detects configured migration tables and ignores them again, automatically... no need for explicitly filtering them out?

If so, that's great!

tobias-93:
@arderyp in fact it is already filtered for the `default` connection, so if you're just using one connection with migrations you can simply remove the filter and enjoy! This PR makes sure this auto-filter is also applied to any additional connection which is defined in the DBAL config.

Pixelshaped:
Having multiple connections and entity managers, I get the `DROP TABLE doctrine_migration_versions;` on `doctrine:migration:diff`. Thanks for your PR, waiting for the release :)

greg0ire:
Thanks @tobias-93 !

cosminsandu:
In what version is this fixed released?

greg0ire:
@cosminsandu AFAIK it's not released yet.

petski:
A [release](https://github.com/doctrine/DoctrineMigrationsBundle/releases) containing this fix will be much appreciated :innocent:

greg0ire:
@petski ask and you shall receive: https://github.com/doctrine/DoctrineMigrationsBundle/releases/tag/3.4.2

:slightly_smiling_face:

faizanakram99:
This fixes doctrine:migrations:diff but not doctrine:schema:validate, that still reports doctrine_migration_versions as a bogus table.

PR #592, opened 2025-02-01
Drop support for ORM 2

greg0ire:


derrabus:
Does supporting ORM 2 mean any additional effort for the bundle? If not, I'd suggest the bundle keeps support for ORM 2 as long as the Migrations library supports it.

greg0ire:
No additional effort, I thought I'd drop the old version before releasing a new major of this bundle. I'm fine with dropping it later as well.

PR #593, opened 2025-02-02
Make all classes final

greg0ire:
This allows us to adapt to upstream changes without that being a breaking change.


SenseException (approved):

stof:
should a 3.x release be done with `@final` applied in phpdoc on them to warn projects that might extend them ?

Btw, some classes should probably also be tagged as `@internal`

PR #606, opened 2025-09-01
Address newly detected PHPStan issues

greg0ire:


greg0ire:
@derrabus @stof the remaing issue about `build_dir` seems to be related to Symfony :thinking: 
Do you maybe know about this?

> Symfony\Component\DependencyInjection\Exception\ParameterNotFoundException: The parameter "doctrine.orm.proxy_dir" has a dependency on a non-existent parameter "kernel.build_dir". Did you mean one of these: "kernel.cache_dir", "kernel.root_dir"?

src/EventListener/SchemaFilterListener.php
```diff
@@ -27,7 +27,11 @@ public function __construct(string $configurationTableName)
     /** @var bool */
     private $enabled = false;
 
-    /** @param AbstractAsset|string $asset */
+    /**
+     * @param AbstractAsset|string $asset
+     *
+     * @phpstan-ignore missingType.generics (a listener should receive all types of assets)
```
stof: The proper way would be to use `AbstractAsset<*>` (where phpstan will then properly check that the logic actually supports any kind of asset)

stof:
@greg0ire `DoctrineMigrationsExtensionTest` is creating its own ContainerBuilder to run the unit tests of the extension. It is not a ContainerBuilder managed by the Kernel. The parameters defined by the Kernel are defined in `DoctrineMigrationsExtensionTest::getContainerBuilder` to "mock" them (sort of), but `kernel.build_dir` is indeed not defined there.

SenseException (approved):

PR #615, opened 2025-10-12
Drop support for old packages

greg0ire:
I was facing difficulties when trying to do up merge up to this branch. This should make things a lot simpler.


dmaicher (approved):
:+1:

PR #621, opened 2025-11-04
Update the Symfony bundle configuration

javiereguiluz:
Without this fix, the doc is not available because `3.5.x` branch doesn't exist in the config file. See https://symfony.com/bundles/DoctrineMigrationsBundle/current/index.html


greg0ire (approved):

greg0ire:
Thanks!

PR #629, opened 2025-11-12
Add service migrations

HypeMC:
A continuation of #524.

Unlike the previous PR, which attempted to support both migrations as services and regular migrations, this one introduces a new `enable_service_migrations` option, which greatly simplifies the overall implementation.

When enabled, Doctrine will fetch migrations from Symfony’s service container.

To use migrations as services, they need to be autoloadable. IMO, the best approach is to place them inside your `src` directory, for example:

```diff
doctrine_migrations:
+   enable_service_migrations: true
    migrations_paths:
-       'DoctrineMigrations': '%kernel.project_dir%/migrations'
+       'App\Migrations': '%kernel.project_dir%/src/Migrations'
```

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -42,9 +42,12 @@ application:
     # config/packages/doctrine_migrations.yaml
 
     doctrine_migrations:
+        # Whether to enable fetching migrations from the service container.
+        enable_service_migrations: false
+
         # List of namespace/path pairs to search for migrations, at least one required
         migrations_paths:
-            'App\Migrations': '%kernel.project_dir%/src/App'
+            'App\Migrations': '%kernel.project_dir%/src/Migrations'
             'AnotherApp\Migrations': '/path/to/other/migrations'
             'SomeBundle\Migrations': '@SomeBundle/Migrations'
 
@@ -234,11 +237,12 @@ Doctrine will then assume that this migration has already been run and will igno
 Migration Dependencies
 ----------------------
 
-Migrations can have dependencies on external services (such as geolocation, mailer, data processing services...) that
-can be used to have more powerful migrations. Those dependencies are not automatically injected into your migrations
-but need to be injected using custom migrations factories.
+Migrations can have dependencies on external services (such as geolocation, mailer or data processing services) to
+enable more advanced behavior. To inject dependencies into your migrations, you must enable loading migrations from
+the service container and register the migrations as services.
 
-Here is an example on how to inject the service container into your migrations:
+If you are using Symfony's default service configuration, migration services are registered automatically
+once placed in the ``src`` directory:
 
 .. configuration-block::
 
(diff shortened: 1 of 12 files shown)


src/DependencyInjection/CompilerPass/RegisterMigrationsPass.php
```diff
@@ -0,0 +1,40 @@
...
+use Doctrine\DBAL\Connection;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\DependencyInjection\Argument\BoundArgument;
+use Symfony\Component\DependencyInjection\Argument\ServiceLocatorArgument;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\DependencyInjection\TypedReference;
+
+/** @final */
```
greg0ire: I think this can be `@internal` and hard `final`.

src/DependencyInjection/Configuration.php
```diff
@@ -34,6 +34,11 @@ public function getConfigTreeBuilder(): TreeBuilder
             ->fixXmlConfig('migration', 'migrations')
             ->fixXmlConfig('migrations_path', 'migrations_paths')
             ->children()
+                ->booleanNode('enable_service_migrations')
+                    ->info('Whether to enable fetching migrations from the container.')
```
greg0ire:
```suggestion
                    ->info('Whether to enable fetching migrations from the service container.')
```

src/MigrationsRepository/ServiceMigrationsRepository.php
```diff
@@ -0,0 +1,67 @@
...
+
+use Doctrine\Migrations\AbstractMigration;
+use Doctrine\Migrations\Exception\MigrationClassNotFound;
+use Doctrine\Migrations\Metadata\AvailableMigration;
+use Doctrine\Migrations\Metadata\AvailableMigrationsSet;
+use Doctrine\Migrations\MigrationsRepository;
+use Doctrine\Migrations\Version\Version;
+use Symfony\Contracts\Service\ServiceProviderInterface;
+
+class ServiceMigrationsRepository implements MigrationsRepository
```
greg0ire: Should be at least `@final`, if not `@internal` + hard final 🤔

src/MigrationsRepository/ServiceMigrationsRepository.php
```diff
@@ -0,0 +1,67 @@
...
+{
+    /** @var ServiceProviderInterface<AbstractMigration> */
+    private $container;
+
+    /** @var AvailableMigration[] */
```
greg0ire:
```suggestion
    /** @var array<string, AvailableMigration> */
```

greg0ire (commented):
Please add documentation in  `docs/index.rst`

src/DependencyInjection/CompilerPass/RegisterMigrationsPass.php
HypeMC: Done

src/MigrationsRepository/ServiceMigrationsRepository.php
HypeMC: I added `@internal` + hard final

src/MigrationsRepository/ServiceMigrationsRepository.php
HypeMC: Done

HypeMC:
@greg0ire I've updated the documentation. Please let me know if it's ok.

greg0ire (approved):

SenseException (approved):

src/DependencyInjection/CompilerPass/RegisterMigrationsPass.php
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        $migrationRefs = [];
+
+        foreach ($container->findTaggedServiceIds('doctrine_migrations.migration', true) as $id => $attributes) {
+            $definition = $container->getDefinition($id);
+            $definition->setBindings([
+                Connection::class => new BoundArgument(new Reference('doctrine.migrations.connection'), false),
+                LoggerInterface::class => new BoundArgument(new Reference('doctrine.migrations.logger'), false),
+            ]);
```
GromNaN, line 32: Does this bundle supports multiple connections? In that case, the appropriate connection must be selected depending on the class namespace.

docs/index.rst
```diff
@@ -42,9 +42,12 @@ application:
     # config/packages/doctrine_migrations.yaml
 
     doctrine_migrations:
+        # Whether to enable fetching migrations from the service container.
+        enable_service_migrations: false
+
         # List of namespace/path pairs to search for migrations, at least one required
         migrations_paths:
-            'App\Migrations': '%kernel.project_dir%/src/App'
+            'App\Migrations': '%kernel.project_dir%/src/Migrations'
```
GromNaN, line 50: The directory should not be required when a namespace is specified. Maybe add a `namespaces` configuration, that is an optional list of PHP namespaces or class names. All migrations from the container are used by default.

src/DependencyInjection/CompilerPass/RegisterMigrationsPass.php
HypeMC: The `DependencyFactory` supports only one connection/em at a time.

docs/index.rst
HypeMC: The directory is still required when generating new migrations so the command knows where to put them.

GromNaN: Oh, all right!

GromNaN (approved):

greg0ire:
Thanks @HypeMC !

PR #644, opened 2025-12-01
Add required return types to MigrationsCollector for Symfony 8 compatibility

webignition:
Method signatures in `MigrationsCollector` require updating to be compatible with base classes and interfaces in Symfony 8.

 - add `void` return type to `MigrationsCollector::collect()`
 - add `void` return type to `MigrationsCollector::reset()`
 - add `string` return type to `MigrationsCollector::getName()`

The `string` and `void` return types were added in PHP 7.0 and 7.1, respectively. As the latest version of this package supports a minimum PHP version of 7.2 I see no backwards compatibility issues or breaking changes.

I'm looking to merge this into `3.7.x` purely to allow for the soonest possible release. Please adjust if this is not what I should be doing.

### DIFF
--- a/src/Collector/MigrationsCollector.php
+++ b/src/Collector/MigrationsCollector.php
@@ -31,7 +31,7 @@ class MigrationsCollector extends DataCollector
     }
 
     /** @return void */
-    public function collect(Request $request, Response $response, ?Throwable $exception = null)
+    public function collect(Request $request, Response $response, ?Throwable $exception = null): void
     {
         if ($this->data !== []) {
             return;
--- a/src/Collector/MigrationsCollector.php
+++ b/src/Collector/MigrationsCollector.php
@@ -88,8 +88,7 @@ class MigrationsCollector extends DataCollector
         return $this->data;
     }
 
-    /** @return void */
-    public function reset()
+    public function reset(): void
     {
         $this->data = [];
     }
--- a/src/Collector/MigrationsCollector.php
+++ b/src/Collector/MigrationsCollector.php
@@ -77,7 +77,7 @@ class MigrationsCollector extends DataCollector
     }
 
     /** @return string */
-    public function getName()
+    public function getName(): string
     {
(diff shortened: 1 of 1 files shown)


greg0ire:
I'm requiring @stof's review because he was not OK with this kind of thing for the bundle. Maybe the right thing here would be to release 4.0.0, but maybe things are different here since we (wrongly) published 3.x versions of this bundle that allow Symfony 8.



I'll attempt to address that last point in https://github.com/doctrine/DoctrineMigrationsBundle/pull/644

@stof, with the above in mind, are you OK with this PR? If not, would you be OK with a PR that would use a trait to add the return types iff `symfony/http-kernel` 8 is in use? If not, I guess we'll just have to release 4.0.0.

src/Collector/MigrationsCollector.php
greg0ire, line 19: The class is annotated `@final`, but that annotation was added in 3.5.0 (see 25b672a21ad1cb8ae91767561dc4ac0f336a38b1)

webignition:
@greg0ire regarding your suggestion of possibly using a trait to add the return types iff `symfony/http-kernel >= 8`, I don't think such conditional application of the return types is strictly necessary. 

In projects of mine being upgraded to symfony 8, I've been preemptively applying a patch to add the return types. This is for projects already on symfony 7.4. Adding return types where none are present in an interface seems to be fine from a syntactic correctness perspective.

greg0ire:
> Adding return types where none are present in an interface seems to be fine from a syntactic correctness perspective.

It is a breaking change for any class extending the class where you add types (but then again the class is annotated with `@final`).

webignition:
Gotcha, thanks for the clarification.

greg0ire:
I just released 4.0.0, so while we wait on a decision regarding 3.x, you may use that version instead.

derrabus:
I think, we can close here. Please use version 4 of the bundle with Symfony 8.

PR #645, opened 2025-12-02
Removed ContainerAwareMigrationFactory usage

### LINKED ISSUE #638: ContainerAwareMigrationFactory configured, but class has been removed
### Bug Report

| Version                                     | 4.0.x-dev

#### Summary

in src/config/services.php the service "doctrine.migrations.container_aware_migrations_factory" is defined with a class "Doctrine\Bundle\MigrationsBundle\MigrationsFactory\ContainerAwareMigrationFactory".
This class has been removed and the service should not removed as well.

It was removed in the previous xml-config, but was re-introduced when that was converted to php in earlier versions, and found it's way in 4.0.x.

#### Current behavior

```
In getDoctrine_Migrations_ContainerAwareMigrationsFactoryService.php line 29:
                                                                                                                                 
  Attempted to load class "ContainerAwareMigrationFactory" from namespace "Doctrine\Bundle\MigrationsBundle\MigrationsFactory".  
  Did you forget a "use" statement for another namespace?    
```

#### Expected behavior

Not fail on the missing service, as well as using the correct new version.

#### How to reproduce

Try installing 4.0.x.
I installed using Symfony 8.0.

Nayte91:
Fixes #638 

## Summary
In SF8 project that uses this bundle, I had an error:

> Attempted to load class "ContainerAwareMigrationFactory" from namespace "Doctrine\Bundle\MigrationsBundle\MigrationsFactory". Did you forget a "use" statement for another namespace? (500 Internal Server Error)

## Explanation
If I understand correctly, this class disappeared in commit d90e9f0c0951cf803e53a448ad44a2f23888e8eb ("Remove support for container-aware migrations (#574)", january), but references were back in commit 5eb080a96af1937d0fefbb92ef43d03c32a2d706 ("Migrate service config from XML to PHP", october). So I think it's an error, when convertir config?

This PR aims to fix that.

### DIFF
--- a/config/services.php
+++ b/config/services.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Symfony\Component\DependencyInjection\Loader\Configurator;
 
 use Doctrine\Bundle\MigrationsBundle\EventListener\SchemaFilterListener;
-use Doctrine\Bundle\MigrationsBundle\MigrationsFactory\ContainerAwareMigrationFactory;
 use Doctrine\Bundle\MigrationsBundle\MigrationsRepository\ServiceMigrationsRepository;
 use Doctrine\DBAL\Connection;
 use Doctrine\Migrations\Configuration\Configuration;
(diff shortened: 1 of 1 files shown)


greg0ire:
Ah yes, it's probably an error when merging up from 3.x to 4.x

greg0ire:
Thanks @Nayte91 !

PR #646, opened 2025-12-02
Fix build

greg0ire:
A lot of the dependencies we use in the code were not declared.

Since the library requires `doctrine/dbal`, it should be fine to require
it in the bundle as well.

Since this is a bundle, it should be fine to require Symfony components.

Since `psr/log` is very widely used, it should be fine to require it.

Note that `symfony/http-kernel` 8 is not allowed yet, because this bundle is
not compatible with it yet.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,11 +25,18 @@
     "homepage": "https://www.doctrine-project.org",
     "require": {
         "php": "^7.2 || ^8.0",
+        "doctrine/dbal": "^2 || ^3 || ^4",
         "doctrine/doctrine-bundle": "^2.4 || ^3.0",
         "doctrine/migrations": "^3.2",
+        "psr/log": "^1 || ^2 || ^3",
+        "symfony/config": "^5.4 || ^6.0 || ^7.0 || ^8.0",
+        "symfony/console": "^5.4 || ^6.0 || ^7.0 || ^8.0",
+        "symfony/dependency-injection": "^5.4 || ^6.0 || ^7.0 || ^8.0",
         "symfony/deprecation-contracts": "^2.1 || ^3",
         "symfony/framework-bundle": "^5.4 || ^6.0 || ^7.0 || ^8.0",
-        "symfony/http-kernel": "^5.4 || ^6.0 || ^7.0"
+        "symfony/http-foundation": "^5.4 || ^6.0 || ^7.0 || ^8.0",
+        "symfony/http-kernel": "^5.4 || ^6.0 || ^7.0",
+        "symfony/service-contracts": "^1.0 || ^2.0 || ^3.0"
     },
     "require-dev": {
         "composer/semver": "^3.0",
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,6 +1,6 @@
 parameters:
     level: 7
-    phpVersion: 80400
+    phpVersion: 80500
     paths:
         - src
         - tests


src/DependencyInjection/Configuration.php
```diff
@@ -22,6 +22,7 @@
 /** @final */
 class Configuration implements ConfigurationInterface
 {
+    /** @return TreeBuilder<'array'> */
```
derrabus, line 25: Why the `'`?

derrabus (commented):
I don't think that we have to declare all of those dependencies. 🤔 I need some time to review this list though.

src/DependencyInjection/Configuration.php
greg0ire: I don't _really_ know, what I do know is that there is a list of possible values here: https://github.com/symfony/symfony/blob/eb611de51bbc7f7466bab42d00a613d79338cc7e/src/Symfony/Component/Config/Definition/Builder/TreeBuilder.php#L19

`'array'` is the default value, which we use, apparently.

SenseException:
Do we really need all those Composer dependencies? Why, e.g., do we explicitly need symfony/console?

greg0ire:
@SenseException great example. Let's examine `symfony/console`. We need it for https://github.com/doctrine/DoctrineMigrationsBundle/blob/3.7.x/src/EventListener/SchemaFilterListener.php

Let's look at the commit message of the commit that introduced it: c85cd25149d82fbc7ae20d03d76f5168a67f39ed


```
    Add support for --complete

    When using orm:schema-tool:update together with --complete, all assets
    not described by the current metadata is dropped.

    This is an issue when using doctrine/migrations, because not using that
    option is deprecated, and because when it is used, the table that holds
    the migrations is dropped as well since it is not described by ORM
    metadata.

    A solution to that is configuring an asset filter that filters out the
    metadata table except when running commands inside the migrations
    namespace.
```

Since `--complete` is becoming mandatory, this class is becoming more and more useful. It looks less and less like something optional. Conversely, migrations is something that is used mostly through the CLI (right?). So what would be so bad about requiring `symfony/console`?

src/DependencyInjection/Configuration.php
stof: because this is about a string literal `'array'`, not about an array type.

stof:
@SenseException those are actually about declaring explicitly the dependencies instead of relying on transitive dependencies, to avoid compatibility issues in case transitive dependencies change their range. They are technically not new dependencies. The code already relies on it.

> Why, e.g., do we explicitly need symfony/console?

because the whole point of that bundle is to expose the CLI commands of `doctrine/migrations` in the Symfony console of the project, wired properly.

composer.json
```diff
@@ -25,10 +25,18 @@
...
         "doctrine/doctrine-bundle": "^2.4 || ^3.0",
         "doctrine/migrations": "^3.2",
+        "psr/log": "^1 || ^2 || ^3",
+        "symfony/config": "^5.4 || ^6.0 || ^7.0 || ^8.0",
+        "symfony/console": "^5.4 || ^6.0 || ^7.0 || ^8.0",
+        "symfony/dependency-injection": "^5.4 || ^6.0 || ^7.0 || ^8.0",
         "symfony/deprecation-contracts": "^2.1 || ^3",
-        "symfony/framework-bundle": "^5.4 || ^6.0 || ^7.0 || ^8.0"
+        "symfony/framework-bundle": "^5.4 || ^6.0 || ^7.0 || ^8.0",
+        "symfony/http-foundation": "^5.4 || ^6.0 || ^7.0 || ^8.0",
```
stof, line 37: does this bundle actually use HttpFoundation ? Where ?

greg0ire: Here: https://github.com/doctrine/DoctrineMigrationsBundle/blob/5a08ff819b34a2931c8d2683e7c8a74569435a28/src/Collector/MigrationsCollector.php#L11

greg0ire:
I extracted the non-controversial parts of this to #651

SenseException (approved):

stof (approved):

PR #657, opened 2025-12-04
Stop using deprecated option

greg0ire:
Since `doctrine/doctrine-bundle` 3.1, `auto_mapping` is deprecated and only accepts false.


greg0ire:
There is another test that uses that deprecated option, but I propose we remove it: https://github.com/doctrine/DoctrineMigrationsBundle/pull/658

derrabus (approved):

PR #667, opened 2026-03-17
[Documentation] Remove annotations from documentation

MrYamous:
IIUC from `composer.json` file, with 4.0.x it's no longer possible to use a version of `doctrine/orm` allowing annotations to define mapping. 
I think we can safely remove this example from doc


greg0ire:
Thanks @MrYamous

PR #682, opened 2026-08-22
Load the configured migration paths when service migrations are enabled

### LINKED ISSUE #680: Project enabling enable_service_migrations removes third party migrations
### Bug Report

| Version                                     | 3.7.0
| Previous Version if the bug is a regression | 3.7.0 ( no enable_service_migrations flag enabled )

#### Summary

The new flag `enable_service_migrations: true` in the Bundle version 3.7.0 can break third party migrations as it not effects the project directory but also vendor directories.

So bundles prepend directories:

```yaml
doctrine_migrations:
    migrations_paths:
        Sulu\Page\Migrations: /Users/alexanderschranz/Documents/Projects/sulu-develop.localhost/vendor/sulu/sulu/packages/page/migrations
        Sulu\Snippet\Migrations: /Users/alexanderschranz/Documents/Projects/sulu-develop.localhost/vendor/sulu/sulu/packages/snippet/migrations
        Sulu\Article\Migrations: /Users/alexanderschranz/Documents/Projects/sulu-develop.localhost/vendor/sulu/sulu/packages/article/migrations
        Sulu\Route\Migrations: /Users/alexanderschranz/Documents/Projects/sulu-develop.localhost/vendor/sulu/sulu/packages/route/migrations
```

but if a project enabled service migrations:

```diff
doctrine_migrations:
    migrations_paths:
        # namespace is arbitrary but should be different from App\Migrations
        # as migrations classes should NOT be autoloaded
        'DoctrineMigrations': '%kernel.project_dir%/migrations'
    enable_profiler: false
+   enable_service_migrations: true
```

#### Current behavior

This removes all third party migrations as they don't define migrations as services.

#### Expected behavior

Such config should not effect third party directories.

#### How to reproduce

See above.

Amoifr:
Fixes #680.

Enabling `enable_service_migrations` replaced the migrations repository with one that only knows the migrations registered as services, so the migrations shipped by third party bundles through `migrations_paths` (for example the ones prepended by Sulu) silently disappeared from `doctrine:migrations:status` and `doctrine:migrations:migrate`.

`ServiceMigrationsRepository` now falls back to the configured migration classes and directories for the migrations that are not registered as services, the same way the default repository of doctrine/migrations loads them (through the configured `MigrationFinder` and `MigrationFactory`, without relying on the internal `FilesystemMigrationsRepository`). Migrations registered as services keep precedence when a class is found on both sides, so projects that list their own `src/Migrations` directory in `migrations_paths` still get the service instances.

While at it, the provided service ids are now read from the keys of `getProvidedServices()` (the values are the service types, `?` for services without a known type).

Covered by a unit test of the fallback and an extension test which boots the container with a service migration and a non-service migration directory.

### DIFF
--- a/src/MigrationsRepository/ServiceMigrationsRepository.php
+++ b/src/MigrationsRepository/ServiceMigrationsRepository.php
@@ -7,6 +7,7 @@ namespace Doctrine\Bundle\MigrationsBundle\MigrationsRepository;
 use Doctrine\Migrations\AbstractMigration;
 use Doctrine\Migrations\Configuration\Configuration;
 use Doctrine\Migrations\Exception\MigrationClassNotFound;
+use Doctrine\Migrations\FilesystemMigrationsRepository;
 use Doctrine\Migrations\Finder\MigrationFinder;
 use Doctrine\Migrations\Metadata\AvailableMigration;
 use Doctrine\Migrations\Metadata\AvailableMigrationsSet;
@@ -16,12 +17,12 @@ use Doctrine\Migrations\Version\Version;
 use Symfony\Contracts\Service\ServiceProviderInterface;
 
 use function array_keys;
-use function array_merge;
-use function class_exists;
 
 /**
- * Loads the migrations registered as services and falls back to the configured migration classes and
- * directories for the other ones (e.g. the migrations shipped by third party bundles).
+ * Loads the migrations registered as services and delegates to the default repository of
+ * doctrine/migrations for the configured migration classes and directories, so that the
+ * migrations which are not registered as services (e.g. the ones shipped by third party bundles)
+ * are available as well.
  *
  * @internal
  */
@@ -39,11 +40,11 @@ final class ServiceMigrationsRepository implements MigrationsRepository
     /** @var MigrationFactory|null */
     private $migrationFactory;
 
+    /** @var MigrationsRepository|null */
+    private $configuredMigrationsRepository;
+
     /** @var array<string, AvailableMigration> */
     private $migrations = [];
-
-    /** @var bool */
-    private $filesystemMigrationsLoaded = false;
 
     /** @param ServiceProviderInterface<AbstractMigration> $container */
     public function __construct(
@@ -64,16 +65,26 @@ final class ServiceMigrationsRepository implements MigrationsRepository
             return true;
         }
 
-        $this->loadMigrationsFromFilesystem();
+        $repository = $this->getConfiguredMigrationsRepository();
 
-        return isset($this->migrations[$version]);
+        return $repository !== null && $repository->hasMigration($version);
     }
 
     public function getMigration(Version $version): AvailableMigration
     {
-        $this->loadMigrationFromContainer($version);
+        $migration = $this->loadMigrationFromContainer($version);
 
-        return $this->migrations[(string) $version];
+        if ($migration !== null) {
+            return $migration;
+        }
+
+        $repository = $this->getConfiguredMigrationsRepository();
+
+        if ($repository === null) {
+            throw MigrationClassNotFound::new((string) $version);
+        }
+
+        return $repository->getMigration($version);
     }
 
     /**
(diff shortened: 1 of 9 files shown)


greg0ire:
> Migrations registered as services keep precedence when a class is found on both sides

How likely is this to happen? Should we throw instead when it does? Silently ignoring a migration sounds pretty bad.

Amoifr:
Thanks for the review! It is actually the documented setup: the docs ask to keep `'App\Migrations': '%kernel.project_dir%/src/Migrations'` in `migrations_paths` (the directory is still needed by `doctrine:migrations:generate`) while the classes in it are registered as services. So every project migration is found on both sides, and nothing is ignored: it is the same class, we just use the container instance (with its injected dependencies) instead of instantiating it a second time through the factory. Throwing there would break every configuration following the docs.

Where the fallback really kicks in is for classes that are not services at all (third party `migrations_paths`, or `migrations` classes listed explicitly). I can reword the PR description if "precedence" sounded like a migration could be dropped.

greg0ire:
> I can reword the PR description if "precedence" sounded like a migration could be dropped.

No need, this was just me being unfamiliar with how this bundle works.

greg0ire:
@HypeMC please review :pray:

src/MigrationsRepository/ServiceMigrationsRepository.php
```diff
@@ -5,31 +5,68 @@
...
+    public function __construct(
+        ServiceProviderInterface $container,
+        ?Configuration $configuration = null,
+        ?MigrationFinder $migrationFinder = null,
+        ?MigrationFactory $migrationFactory = null
+    ) {
+        $this->container        = $container;
+        $this->configuration    = $configuration;
+        $this->migrationFinder  = $migrationFinder;
+        $this->migrationFactory = $migrationFactory;
```
greg0ire, line 59: Given the projects' current constraints, I think you could be using constructor property promotion here.

src/MigrationsRepository/ServiceMigrationsRepository.php
```diff
@@ -60,9 +99,53 @@ private function loadMigrationFromContainer(Version $version): void
...
         }
 
         $this->migrations[$id] = new AvailableMigration($version, $this->container->get($id));
     }
+
+    /**
+     * Registers the migrations found in the configured classes and directories which are not
+     * registered as services, the way the default repository of doctrine/migrations does.
+     */
+    private function loadMigrationsFromFilesystem(): void
```
greg0ire: I think I would expect `ServiceMigrationsRepository` to delegate to another implementation of `MigrationsRepository`, or for these 2 implementations to be wrapped in a `ComponsiteMigrationsRepository`. Here, it sound like you are copying logic maintained in the library to the bundle. So if we find a big in that logic, we would have to fix it in 2 places.

src/MigrationsRepository/ServiceMigrationsRepository.php
Amoifr: I'd love to, but `3.7.x` still requires `php ^7.2 || ^8.0` (the CI runs 7.2 to 8.4), so constructor promotion is out of reach on this branch. Happy to switch if you'd rather target a newer branch.

src/MigrationsRepository/ServiceMigrationsRepository.php
Amoifr: Agreed, and done in the second commit: the repository now delegates to the library's `FilesystemMigrationsRepository`, built from the `Configuration` the same way `DependencyFactory::getMigrationRepository()` does, and only keeps the service resolution. The one twist is a tiny `MigrationFactory` decorator (`ServiceMigrationFactory`) handed to that repository: the project migrations are both registered as services and listed in `migrations_paths` (as documented), so without it the library would instantiate them a second time through `DbalMigrationFactory` and blow up on their injected dependencies. One thing to confirm on your side: `FilesystemMigrationsRepository` is marked `@internal` in doctrine/migrations, which is why I had avoided it. Fine for the bundle, or would you prefer the library to expose it?

src/MigrationsRepository/ServiceMigrationsRepository.php
greg0ire: Oh, right, I took a look at the wrong branch, sorry.

src/MigrationsRepository/ServiceMigrationsRepository.php
greg0ire: Why does `FilesystemMigrationsRepository` need to be instantiated in this bundle? Can't `ServiceMigrationsRepository` delegate to the instance is that would usually be created if `enable_service_migrations` was `false`?

HypeMC:
@greg0ire Sorry for the late reply, this one flew under my radar. I made PR #683 with a different approach. It's similar to this one, but adds a `CompositeMigrationsRepository` as you suggested and reuses the existing `FilesystemMigrationsRepository` instance.

Amoifr:
Thanks @HypeMC, #683 is the better shape: the `CompositeMigrationsRepository` greg0ire suggested plus reusing the existing `FilesystemMigrationsRepository` instance answers the open question here in one move. Happy to see it supersede this PR, feel free to close this one in its favor. And thanks @greg0ire for the review, the delegation direction clearly was the right call.

stof:
Closing in favor of #683

